PR #1051, opened 2014-12-30
[MOD] Select files/dirs in the project view within the whole row with a ...

dirkk:
...left click as well.

I do think this is much more practical, because otherwise I keep missing to click the `a` folder...

screenshot-area-2014-12-30-125415


ChristianGruen:
Sorry, I've been missing this one. I chose a slightly different solution to still allow multi-selections via the Shift key.

PR #1068, opened 2015-01-31
Update Hungarian.lang

KissDaniGH:


ChristianGruen:
…reliable as always ;)

Issue #1060, opened 2015-01-20
Messy latest stable WAR file


kristiank:
Hello!

Background: I have a functioning install of the latest official release, but because of the dba and other features (mainly users.xml), so I wanted to try version 8.0.

So I downloaded the latest stable from [here](http://files.basex.org/releases/latest/BaseX80-20150118.150705.war) and deployed it in Tomcat 7.0.52.

Many issues arised:

* When un-commenting the ``<context-param>`` in WEB-INF/web.xml, it turns out the web.xml is malformed and contains several opening elements of ``<context-param>`` without matching ending elements. I erased the opening elements and get the servlet started.

* When opening [my localhost](http://localhost:8080/BaseX80-20150118.150705/) I noticed the graphics and style is missing -- turns out the whole static folder is missing in the war file! I copy the two files from my old 7.9 installation and the start-page looks fine.

* When opening the [dba](http://localhost:8080/BaseX80-20150118.150705/dba) I get a "HTTP Status 404 - No function found that matches the request." message.

Probably I should change my workflow and download the latest code from git and compile it, but that would require a too steep learning curve. Hopefully a new latest stable can be generated by the BaseX team instead. Thank you!

ChristianGruen:
Thanks for the helpful hints. Our aim is to make Version 8.0 as easily usable as 7.9.

kristiank:
I will try using Git and Eclipse. Your documentation seems to make it less steep than I thought.

kristiank:
... but wait, if I simply do an automated build from the latest code, shouldn't I get the same result as your latest stable has?

ChristianGruen:
> ... but wait, if I simply do an automated build from the latest code, shouldn't I get the same result as your latest stable has?

Yes, it should work out like this.

I have fixed Issue 1. Regarding Issue 2, the new `rest.xqm` only includes a redirection to DBA (but maybe I will reinclude the old start page). I would be interested what's the reason for Issue 3.

ChristianGruen:
The start page, and the static folder, is now reincluded in the distributions. On Issue 3: Did you manage to successfully start DBA? If no, do you have some clue what could be the problem?

kristiank:
[...]

PR #1075, opened 2015-02-07
Changed redirect to a relative address

kristiank:
When an absolute address was used, the redirection possibly stripped the base path away. Especially using BaseX as a Tomcat webapp, the redirection was broken. Instead of the correct redirection:
``http://localhost:8080/webapp-name/dba`` -> ``http://localhost:8080/webapp-name/dba/databases``
this incorrect was produced:
``http://localhost:8080/webapp-name/dba`` -> ``http://localhost:8080/dba/databases``

 Using a relative address should work *always*.

This solves a problem partially mentioned in #1060

### DIFF
--- a/basex-api/src/main/webapp/dba/util.xqm
+++ b/basex-api/src/main/webapp/dba/util.xqm
@@ -17,7 +17,7 @@ declare
   %rest:path("dba")
 function _:redirect(
 ) {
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Makes sense; thank you!

PR #1089, opened 2015-03-05
Two more HOFs: `hof:take-while($seq, $pred)` and `hof:scan-left($seq, $start, $f)`.

LeoWoerteler:
* `hof:take-while($seq, $pred)` is equivalent to:

   ```xquery
   declare function hof:take-while($seq, $pred) {
     if(empty($seq) or not($pred(head($seq)))) then ()
     else (
       head($seq),
       hof:take-while(tail($seq), $pred)
     )
   };
   ```
 * `hof:scan-left($seq, $start, $f)` is equivalent to:

   ```xquery
   declare function hof:scan-left($seq, $acc, $f) {
     if(empty($seq)) then $acc
     else (
       $acc,
       hof:scan-left(tail($seq), $f($acc, head($seq)), $f)
     )
   };
   ```

### DIFF
--- a/basex-core/src/main/java/org/basex/query/func/Function.java
+++ b/basex-core/src/main/java/org/basex/query/func/Function.java
@@ -970,7 +970,14 @@ public enum Function {
 
   /** XQuery function. */
   _HOF_SORT_WITH(HofSortWith.class, "sort-with(items,lt-fun)",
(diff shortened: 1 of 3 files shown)


ChristianGruen:
To be added to http://docs.basex.org/wiki/Hof_Module …

```xquery
hof:scan-left(1 to 10, 0, function($a, $b) { $a + $b }),

hof:take-while(
  let $n := 100000000
  for $i in 1 to $n
  return ((1 to $n)[. = $i])[1],
  function($x) { $x <= 5 }
)
```

PR #1093, opened 2015-03-11
Retain system properties when starting in service mode

davidmathei:
When starting in service mode, system properties got lost as the key of the property was not copied into the arguments of the new process.


ChristianGruen:
Thanks!

Issue #1104, opened 2015-03-23
fn:random-number-generator() is not referentially transparent


LeoWoerteler:
## Query
```xquery
let $rng := fn:random-number-generator()
return $rng?next()?number eq $rng?next()?number
```
## Returns:
`false` (most of the time)

## Expected:
`true`

PR #1107, opened 2015-03-23
GH-1104: Remove mutable state from `fn:random-number-generator([$seed])`

LeoWoerteler:


ChristianGruen:
Nice to have you back on the stage… Thanks ;)

PR #1111, opened 2015-03-25
Do not ignore $CLASSPATH additions

JensErat:
BaseX sets `-cp` when starting java to include its own classes, but thus overwrites the `$CLASSPATH` environment variable (java ignores it when `-cp` is set). This does not leave the developer the choice to use environment variables for loading additional helper classes, which sometimes is easier and cleaner compared to linking into the BaseX `lib` folder (or even isn't possible if running on a shared host without write privileges to `/usr`).

With the patch, for example Apache batik could be loaded using

    export CLASSPATH=/usr/share/java/batik.jar
    basexhttp


ChristianGruen:
Agreed; thanks for the pull request.

PR #1128, opened 2015-04-20
URI resolving mechanism concept for xquery imported modules.

hovi:
This is not finished work and serves only as example for further discussion, more information in mailing list:
https://mailman.uni-konstanz.de/pipermail/basex-talk/2015-April/thread.html#8613


ChristianGruen:
Thanks for your efforts! We will review this soon.

PR #1131, opened 2015-04-21
Moved `ValueBuilder` to `org.basex.query.value`.

LeoWoerteler:


ChristianGruen:
"Für gut befunden", as Germans say.

PR #1146, opened 2015-05-23
Syntax highlighting in Queries editor

jb8748:
It's not to the level of sophistication of the GUI but I've used CodeMirror to implement basic syntax highlighting in the Editor portion of the Queries screen in DBA.

This pull request has the required changes to the BaseX files.

Additionally you need to have the CodeMirror files. As I'm not sure how you include (or would include) files from other projects in BaseX so I haven't added them here.

I put a folder in 'files' called 'codemirror'. Inside there:

./lib/codemirror.css
./lib/codemirror.js
./mode/xquery/xquery.js
./LICENSE

syntax-highlighting

### DIFF
--- a/basex-api/src/main/webapp/dba/files/editor.js
+++ b/basex-api/src/main/webapp/dba/files/editor.js
@@ -4,6 +4,10 @@ function openQuery() {
     null,
     function(req) {
       document.getElementById("editor").value = req.responseText;
+      //The next line sends a 'change' event to all listeners to the 'editor'
+      //This allows the syntax highlighter to hear that the text area value has changed
+      //Programatic changes to the value don't send events automatically
+      document.getElementById("editor").dispatchEvent(new Event("change",{ bubbles: false, cancelable: false }));
     },
     function(req) {
       setError('Query could not be opened.');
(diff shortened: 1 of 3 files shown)


ChristianGruen:
Thanks a lot for your pull request! I'll look into this soon.

ChristianGruen:
I really like your idea to add highlighting to the editor; it improved usability a lot. I don't see any problems in adding the CodeMirror code to our repository, even if it's 3 times the size of the current code (well, it's all tiny anyway).

Here are some issues that might need to be fixed before merging the code:

* The 'Realtime' mode doesn't work anymore
* Before (at least on Windows) it was possible to switch to the "Run" button by typing Shift+Tab. These keys now seem to be absorbed by the editor.
* Maybe we could use XML highlighting for the 'output' textarea? One side effect would be that the editor and output textareas would then have the same size again
* A minor one: With browsers like Chrome, the size of the textareas could be resized. That's not possible anymore when using CodeMirror. However, most people can surely live with it, as it would be nicer anyway if the textareas were automatically vertically resized to fill the full window.

Would you be interested in having a look at it?

jb8748:
Thank you for your comments. I certainly find the editor much more useful now.

On the specifics:
1) Ok, I don't think realtime has ever worked for me so I didn't realise this broke it elsewhere.
2) I think there's a way to do key mapping so maybe this is fixable
3) I've already done that so I can put that in the next update
4) Yes I saw this, it's the same in Safari. Having done some work recently on single-page apps I decided it wasn't worth fixing for the effort. But let me have a think.

I can't promise it will be quick but I'll have a look at all these and update the pull.

ChristianGruen:
I was stupid enough to already merge your pull request into the master branch.. Which is why you'll probably have to add a new pull request. Sorry.

PR #1147, opened 2015-05-23
Explicitly set Content-Type when making XHR request from DBA

jb8748:
Explicitly setting Content-Type to text/plain in the request function of js.js. When it is not set, Safari (testing on 7.1.6 Mac) defaults to 'application/xml' and so BaseX refuses to accept anything that isn't XML. This can be seen when trying to save an xquery.

(Firefox defaults to 'text/plain' so there shouldn't be any change in behaviour there.)

I think I've checked for all current cases where this function is called as POST with data (the only time it'll be a problem) and text/plain works for all.


ChristianGruen:
Looks completely reasonable; thanks in advance!

PR #1153, opened 2015-05-29
Correct variable name in drop() function in Optimize

jb8748:
Pressing Optimize on the databases screens fails with an error that $names is not defined.

In optimize.xqm the function string call to util:update contains the variable name $names when the map variable is called $n. This now matches the function in databases.xqm.


ChristianGruen:
+

PR #1154, opened 2015-05-31
Syntax highlighting in Queries (2)

jb8748:
This pull request has the required changes to the BaseX files.

Additionally you need to have the CodeMirror files. As I'm not sure how you include (or would include) files from other projects in BaseX so I haven't added them here.

I put a folder in 'files' called 'codemirror'. Inside there:

./lib/codemirror.css
./lib/codemirror.js
./mode/xquery/xquery.js
./mode/xml/xml.js
./LICENSE

In this revised version:
- syntax highlighting of both the editor and result (XQuery and XML respectively)
- the realtime mode should work again
- resizing of the areas by dragging proved too difficult but I've implemented a function to expand them to the available screen height
- I've added a keybinding on the editor (CTRL-ENTER and CMD-ENTER) that will run the query. It's not the SHIFT-TAB you had before but it does match the GUI (and means you can use tabs in the editor)

### DIFF
--- a/basex-api/src/main/webapp/dba/files/editor.js
+++ b/basex-api/src/main/webapp/dba/files/editor.js
@@ -4,7 +4,9 @@ function openQuery() {
     null,
     function(req) {
       document.getElementById("editor").value = req.responseText;
-      document.getElementById("editor").dispatchEvent(new Event("change",{ bubbles: false, cancelable: false }));
+      var evt = document.createEvent("HTMLEvents");
+      evt.initEvent("change",false,false);
+      document.getElementById("editor").dispatchEvent(evt);
     },
     function(req) {
       setError('Query could not be opened.');
@@ -76,16 +78,8 @@ function queryExists() {
   return false;
 };
 
-function loadCodeMirrorCSS() {
-    var link = document.createElement("link");
-    link.setAttribute("rel", "stylesheet");
-    link.setAttribute("type", "text/css");
-    link.setAttribute("href", "files/codemirror/lib/codemirror.css");
-    document.head.appendChild(link);
-}
-
 function loadCodeMirror() {
-  if (CodeMirror) {
+  if (CodeMirror && dispatchEvent) {
     //Now replace the editor and result areas
     var editorArea = document.getElementById("editor");
     var codeMirrorEditor = CodeMirror.fromTextArea(editorArea, {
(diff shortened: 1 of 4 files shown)


ChristianGruen:
I really like the support of Ctrl-/Cmd-Enter and the utilization of the available screen height. And it works fine with Chrome 42! Unfortunately, though, I can't edit queries anymore with other browsers (Firefox 38, Opera 12, IE 11)… Could you possibly try and see if it works with Firefox on your system?

jb8748:
Hmm. When you say you can't edit queries do you mean the edit area doesn't work? Or you can't open and then save a query?

I've just tried Opera (Mac 29) and it's ok.
I don't have IE.
FireFox 38 is misbehaving - giving me an almost completely grey editor. Is this what you're seeing?

I think the problem with FireFox is that I'm loading the CSS file via script and it's getting confused. I did this so that I didn't have to change the tmpl.xqm file and load the CodeMirror CSS on every page. However, if you're happy to do this then I can update the pull request.

Or - should I adjust tmpl.xqm to allow for another entry in the options map for stylesheets?

Thoughts welcomed.

ChristianGruen:
> However, if you're happy to do this then I can update the pull request.
> Or - should I adjust tmpl.xqm to allow for another entry in the options map for stylesheets?

Yes, you are invited to extend the XQuery code. The DBA code was developed in a pretty short time span, and I already regret that I initially moved all the Javascript code in a single `js.js` file..

I have attached the Win 8.1 screenshots for Firefox, Opera 12 and IE (Opera 29 works fine on my system, too; surely, because it's also based on the Blink renderer). IE does correctly display the highlighting, but querying fails with the code of the pull request (the message "Please wait..." does not disappear, and the result is not shown anymore). Maybe I can fix this by myself.

firefox
opera
ie

jb8748:
Thank you for the update. I've already made the changes to tmpl.xqm so you can now pass a set of strings in for adding additional CSS and another set for additional scripts. That fixes FireFox for me and I hope will fix Opera 12 for you.

The issue with IE is a different one - I've fired up my VM and isolated the issue - I've just got to work out how to fix it.

I'll update the code soon.

jb8748:
Ok, so I think I've managed to solve all the things you've found - although I can't test in Opera 12 so please do.

I don't know how old a version of IE will be supported but it seems to be working for me in 11.

ChristianGruen:
Good news… All works like a breeze now!

PR #1162, opened 2015-06-15
[FIX] Allow global read lock with concurrent write lock

JensErat:
BaseX ran into a deadlock when a global read lock with a
concurrent local write lock was requested.

### DIFF
--- a/basex-core/src/main/java/org/basex/core/locks/DBLocking.java
+++ b/basex-core/src/main/java/org/basex/core/locks/DBLocking.java
@@ -138,7 +138,9 @@ public final class DBLocking implements Locking {
       }
       // global read locking
       if(read == null) {
-        while(localWriters > 0) {
+        while(localWriters > 0 &&
+            // We're the only writer, allow global read lock anyway
+            !(1 == localWriters && !(null == write || write.isEmpty()))) {
           try {
             globalLock.wait();
           } catch(final InterruptedException ex) {
--- a/basex-api/src/main/webapp/dba/settings/settings.xqm
+++ b/basex-api/src/main/webapp/dba/settings/settings.xqm
@@ -9,6 +9,8 @@ import module namespace Request = 'http://exquery.org/ns/request';
 import module namespace cons = 'dba/cons' at '../modules/cons.xqm';
 import module namespace tmpl = 'dba/tmpl' at '../modules/tmpl.xqm';
 
+declare option query:write-lock 'settings';
+
 (:~ Top category :)
 declare variable $_:CAT := 'settings';
 
--- a/basex-core/src/test/java/org/basex/core/LockingTest.java
+++ b/basex-core/src/test/java/org/basex/core/LockingTest.java
@@ -64,6 +64,21 @@ public final class LockingTest extends SandboxTest {
   }
 
   /**
+   * Single thread acquires both global read lock and a single write lock.
+   * @throws InterruptedException Got interrupted.
+   */
+  @Test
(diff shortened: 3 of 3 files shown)


LeoWoerteler:
Why should this be allowed? Couldn't the reader perform a dirty read of the locked object?

JensErat:
Not in case reader and writer are actually the same thread, ie. a query requesting global read and local write locks at the same time.

Two different queries will still not be allowed to do so, as the already existing test `LockingTest#globalReadLocalWriteLockingTest` is ensuring.

basex-core/src/main/java/org/basex/core/locks/DBLocking.java
```diff
@@ -138,7 +138,9 @@ public void acquire(final Proc pr, final StringList read, final StringList write
       }
       // global read locking
       if(read == null) {
-        while(localWriters > 0) {
+        while(localWriters > 0 &&
+            // We're the only writer, allow global read lock anyway
+            !(1 == localWriters && !(null == write || write.isEmpty()))) {
```
ChristianGruen: I'm trying to understand what's going on in here.. So the loop is either interrupted if:

1. `localWriter = 0`, or
2. `localWriter = 1` and:
  1. the write lock is global, or
  2. there are no local write locks

Did I get it right? If yes, the code could possible be simplified a bit as follows:

```java
while(localWriters > 0 && (localWriters > 2 || write == null || write.isEmpty()))
```

JensErat: Yes it could be, but I decided to skip the "normalization" and stick closer to the semantics: "There's only one writer, and we're this one".

ChristianGruen: Well, it was quite a challenge for me to correctly decode the negated conditions ;) But as long as it's correct, it should be fine.

By the by, I am indeed afraid that I probably got it wrong. This is my updated attempt… The loop will be interrupted (e.g. `globalLock.wait()` will not be called anymore) if:

1. `localWriters` = `0`, or
2. `localWriters` = `1` and if there is at least one local write lock

Do you agree? I'm asking to ensure that the code is really correct, and that all cases are correctly covered.

JensErat: Yes, this seems reasonable to me. In other words:

1. There are _no_ `localWriters` at all (this was already allowed)
2. There is _exactly one_ `localWriter`, and the current query is a `localWriter`, so there cannot be any other.

ChristianGruen: Ok, thanks. For my own reference, I'm just listing the cases that are implied by the conditions:

**Loop:**
* `localWriters` > `1`
* `localWriters` = `1`, global write
* `localWriters` = `1`, no local writes

**Stop:**
* `localWriters` = `0`
* `localWriters` = `1`, local writes

JensErat: Yes, that's it.

**Loop (wait for resources to be released):**
* `localWriters` > `1` (there is somebody else writing somewhere)
* `localWriters` = `1`, global write (we did not specify a local write lock, cannot be us)
* `localWriters` = `1`, no local writes (we do not write, cannot be us)

**Stop (resources are available, continue:**
* `localWriters` = `0` (no writers at all, fine anyway)
* `localWriters` = `1`, local writes (there is exactly one writer, and we are this one)

PR #1170, opened 2015-07-07
[MOD] Query should be sent as POST parameter 

dirkk:
Otherwise, the entity body can easily be too large.

HTTP Code 413 "Entity Too Large" is returned in these cases, the limit is even quite low normally.


ChristianGruen:
Thanks in advance! I'll check it pretty soon.

ChristianGruen:
I didn't push this yet, because it requires some (probably minor) changes for the query feature in the Database panel. Could you have a look at this?

dba

dirkk:
Done

PR #1196, opened 2015-09-17
Use original casing of archive name.

vincentml:
Hi Christian,

Thanks for adding this feature! It seems to work well. I noticed that the archive file name is being converted to lower case in the path. This pull request is to use the original casing of the archive file name. 

Vincent


ChristianGruen:
Makes complete sense. Thanks for the pull request, very appreciated!

PR #1205, opened 2015-10-21
Proposal for passing authorization ahead to restxq when scheme isn't known

dcore94:
Depending on a boolean property CUSTOMAUTH the authorization could be passed on to the application code (RESTXQ) instead of raising an exception.
This allows for custom authorization handling (use case: Authorization: Bearer XXXYYYZZZ).

### DIFF
--- a/basex-api/src/main/java/org/basex/http/HTTPContext.java
+++ b/basex-api/src/main/java/org/basex/http/HTTPContext.java
@@ -119,7 +119,7 @@ public final class HTTPContext {
     final String[] ams = Strings.split(value, ' ', 2);
     final AuthMethod am = StaticOptions.AUTHMETHOD.get(ams[0]);
     if(am == null){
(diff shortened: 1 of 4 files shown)


ChristianGruen:
Sorry for ignoring your pull request for quite some time now… Is it still interested in having this in the core? – I thought about allowing an additional `CUSTOM` value for the `AUTHMETHOD` option, would this help?

dcore94:
It is of use still. 
After your possible addition to the core, it should be possible to set AUTHMETHOD = CUSTOM in a .basex file obtaining that the validation of the Authorization header will be delegated to RESTXQ code instead of passing through the default Basex filters. Is it right?
In this case I'd say it's what the pull request was aiming at
Thanks a lot for putting another thought on this!

ChristianGruen:
Thanks for your swift feedback. Could you have a look at my latest commit, assign `Custom` to `AUTHMETHOD`, and see if it works as expected?

Issue #928, opened 2014-04-13
Don't use java.net.URI to validate URIs


dimitarp:
BaseX uses java.net.URI for namespaces. However, java.net.URI is unfortunately not compliant with RFC 3986. Thus the following is considered invalid by BaseX:

```xml
<a xmlns="x:"/>
```

ChristianGruen:
I think I'll close this (sounds like a lot of effort.. ;).

ChristianGruen:
Copied from #1215:

> Consider the following query
> ```xquery
> resolve-uri("file:///b.xml"),
> resolve-uri("a.xml","file:///b.xml")
> ```
> the result is
> ``` 
> file:///b.xml
> file:/a.xml
> ```
> It would be nice if both results had the same number of slashes.

PR #1233, opened 2015-12-26
[FIX] GH-928: RFC 3986 URI parser

dimitarp:
Fixes the original problem from GH-928 (i.e. correct URI resolving is not included).


LeoWoerteler:
:+1: nice!

ChristianGruen:
Thanks a lot!

PR #1234, opened 2015-12-26
Some Code Fixes

dimitarp:
Some small code fixes (feel free to choose what to merge).
Merry Christmas to all BaseX devs and users! :)


ChristianGruen:
I decided to keep the generated files, as it simplifies code execution for people without maven and git (sounds surprising, but that’s still an issue ;). Thanks again!

dimitarp:
What about f4c7014? I spoke with @LeoWoerteler about how to run the W3C XQuery test suite - one should use the class ``org.basex.tests.w3c.QT3TS``. The deleted test classes should no longer be used (actually the code there is broken) and since it is not documented how to run the test suite, they bring some confusion in the otherwise clean code ;)

ChristianGruen:
Makes sense! In that case, I think, we should only remove classes in the `basex-tests/src/test/java/org/basex/qt3ts` directory (i.e., keep classes such as [FNCryptoTest](https://github.com/BaseXdb/basex/blob/master/basex-tests/src/test/java/org/basex/FNCryptoTest.java), [MultipleAddTest](https://github.com/BaseXdb/basex/blob/master/basex-tests/src/test/java/org/basex/local/multiple/MultipleAddTest.java)); I am still running the other tests at irregular intervals.

PR #1251, opened 2016-01-20
New QueryContext and QueryResources contructors to shallow copy resources and support transform expressions in parallel queries.

james-jw:
In further testing of xq-promise, which attempts to parallelize non Updating XQuery expressions. I have found one issue. Namely, transformations such as the following do not work in parallel callbacks.

```xquery
let $base-node := <node />
let $func := function ($item, $i) {
   copy $out := $item
   modify (
     rename node $out as 'test-transform',
     insert node element value {$i}
   )
   return $out
}
return p:fork-join(
  for $i in (1 to 100) return
  p:defer($func(?, $i), $base-node)
) 
```

In particular the error observed is `ArrayOutOfBounds`. It pertains to the `Updates` variable contained on the QueryContext.resources object.

The xq-promise module as it stands currently creates a new QueryContext from the parent query during a fork or fork-join operation using the QueryContext(QueryContext qc) contstructor. This is done in order to try and isolate the threads as much as possible. The one issue I have found however is this constructor simply copies the reference to the `resources` object of the parent QueryContext, which includes this Updates list.

Since the update list is mutable and will be changed during a transformation, each forked child QueryContext needs to have its own resource and in particular Update lists or else they will stomp on each other.

The idea behind this merge is to provide a QueryContext constructor which shallow copies the parent QueryContext's resources into its own set of ArrayLists and also maintains its own `Updates` list. With this change, the above example and other transform queries works as expected. 

Going forward, with a little further understanding of how the Updates are managed, I could see database updates being parallelize as well, but with some extra effort. Essentially the fork-join code would need to merge the individual threads update lists after they returned. Haven't tried that yet, but its maybe workable? Thoughts?

For now the xq-promise module throws an exception if you try and add any database updating expressions.

I am curious to hear all your thoughts.

Cheers!

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryContext.java
+++ b/basex-core/src/main/java/org/basex/query/QueryContext.java
@@ -169,6 +169,20 @@ public final class QueryContext extends Proc implements Closeable {
   }
 
   /**
+   * Construct for use in a fork-join query
+   * @param qcParent
+   * @param async
+   */
+  public QueryContext(final QueryContext qcParent, boolean async ) {
+    this(qcParent.context, qcParent);
+    listen = qcParent.listen;
+    if(async) {
+      resources = new QueryResources(this, qcParent.resources);
+    } else {
+      resources = qcParent.resources;
+    }
+  }
+  /**
    * Constructor.
    * @param context database context
    * @param parent parent context (can be {@code null})
--- a/basex-core/src/main/java/org/basex/query/QueryContext.java
+++ b/basex-core/src/main/java/org/basex/query/QueryContext.java
@@ -169,9 +169,11 @@ public final class QueryContext extends Proc implements Closeable {
   }
 
   /**
-   * Construct for use in a fork-join query
-   * @param qcParent
-   * @param async
+   * Construct context for use in a parallel query. Creates shallow clone
+   * of resources in order to mitigate multi threading array syncronization issues.
+   * @param qcParent parent context
+   * @param async If true the resources will be shallow copied. If false,
+   * operates the same as if calling the standard constructor: QueryContext(final QueryContext qc) 
    */
   public QueryContext(final QueryContext qcParent, boolean async ) {
     this(qcParent.context, qcParent);
--- a/basex-core/src/main/java/org/basex/query/QueryResources.java
+++ b/basex-core/src/main/java/org/basex/query/QueryResources.java
@@ -63,6 +63,11 @@ public final class QueryResources {
     this.qc = qc;
   }
 
+  /**
+   * Contructor. Shallow copies the resources for use in paralle queries.
+   * @param qc query context
+   * @param qr resources to clone
+   */
   @SuppressWarnings("unchecked")
   QueryResources(QueryContext qc, QueryResources qr) {
     this.qc = qc;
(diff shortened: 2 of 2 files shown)


ChristianGruen:
Thanks! I’ll check it out in a while.

james-jw:
Definitely not rush! Thanks again.

ChristianGruen:
Hi James, thanks again for your pull request! I was indeed surprised how easy it is to even parallelize requests that access external resources, such as XML documents.

This is what I finally did:

* I moved references to the update structures to the QueryContext class, and I changed the way how these structures are initialized. This way, it cannot happen anymore to have concurrent changes of the required references.
* I kept the single pool of query resources, because one important property of XQuery is that all documents with a specific URI must have the same internal reference. As a result, `doc('x.xml') is doc('x.xml')` is expected to return true, and it should also return true if the functions are evaluated in different threads:

```xquery
import module namespace p = 'https://github.com/james-jw/xq-promise' at 'xq-promise.xqm';
(: doc(<x>x.xml</x>) looks stupid, but it prevents opening of x.xml at compile time :)
let $func := function() { doc(<x>x.xml</x>) }
let $nodes := p:fork-join( (1 to 10000) ! p:defer($func) )
(: expected to yield true :)
return every $n in $nodes satisfies $n is $nodes[1]
```

Looking forward to your testing feedback!

james-jw:
Christian, This is great news! I will hopefully have time tomorrow to get testing started on this!

Ill keep you posted!

james-jw:
Works like a charm! 

I think BaseX users, once we confirm its safety, will really enjoy this. Xavier is already seeing encouraging results in his linguistics research (testing)

Ill keep you posted on anything I find. 

On a side note, I was able to completely rewrite the pattern in XQuery. No more XqDeferred.java class. Basically the only thing left in Java is the RecusiveTask and ForkJoinPool management. I essentially use closures like Rob Stapper demonstrated with his custom data types. 

The checking was:
-750 lines and only +108

Check out the implementation in the xq-promise.xqm file if your curious, its short.

With that in mind, I don't see a huge obstacle to simply adding a fork-join function as a module to BaseX once your confident in it. Even as a spec function in XQuery 3.1.x

The thing is, this fork-join function doesn't require the promise pattern. It would essentially fork-join any functions, but with the pattern it makes it even more powerful/efficient. Thoughts?

james-jw:
Actually the last example you posted is not working, but with or without fork-join. Maybe a typo? I get this error:

```
Resource 'C:/Users/SESA221445/workspace/xq-promise/tests/x.xml' does not exist.
```

```xquery
let $func := function() { doc(<x>x.xml</x>) }
  return $func()
```

ChristianGruen:
> -750 lines and only +108

Awesome! And cool to see that we already got some new testers.

> The thing is, this fork-join function doesn't require the promise pattern. 

Oh, right, good point. I’ll have some more thoughts on that.

> Resource 'C:/Users/SESA221445/workspace/xq-promise/tests/x.xml' does not exist.

Sorry, I forgot to add that you’ll need an `x.xml` file to make it work (`<x>x.xml</x>` will be resolved to the string `x.xml`, and this file will then be retrieved from disk).

PR #1257, opened 2016-01-31
[HUN] update

KissDaniGH:


KissDaniGH:
One string left. New commit is coming.

PR #1262, opened 2016-02-09
fix handling of undefined key events

klavs:
From [KeyEvent docs](https://docs.oracle.com/javase/7/docs/api/java/awt/event/KeyEvent.html#VK_QUOTE):
> For key typed events, the getKeyCode method always returns VK_UNDEFINED.

> Not all keyboards or systems are capable of generating all virtual key codes. No attempt is made in Java to generate these keys artificially.

When I press `"` key the `KeyEvent e` returns `e.getKeyCode() == VK_UNDEFINED` and `e.getKeyChar() == '"'` also `BaseXKeys` returns `NEXTPAGE.is(e) == true` which is false.

`NEXTPAGE` is defined as `NEXTPAGE(NO_MOD, VK_PAGE_DOWN, SHIFT)` so `e.getKeyChar()` should never be compared to `VK_PAGE_DOWN`.

There could be one problem with this fix: `GOBACK2(VK_BACK_SPACE, NO_MOD)` sets `key = NO_MOD`, which equals to `key = 0`.

### DIFF
--- a/basex-core/src/main/java/org/basex/gui/layout/BaseXKeys.java
+++ b/basex-core/src/main/java/org/basex/gui/layout/BaseXKeys.java
@@ -12,6 +12,7 @@ import java.awt.event.*;
  * @author BaseX Team 2005-16, BSD License
  * @author Christian Gruen
  * @author Leo Woerteler
+ * @author Klavs Prieditis
  */
 public enum BaseXKeys {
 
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Thanks for the pull request; I'll check it out soon.

ChristianGruen:
Your code has been merged, thanks again. I simply removed the `GOBACK2` key binding, as it didn’t work as as expected, even before your fix.

PR #1264, opened 2016-02-11
Compare Strings using equals()

AymanDF:
This pull request is focused on resolving occurrences of Findbugs rule ES_COMPARING_STRINGS_WITH_EQ - “Bad practice - comparison of String objects using == or != ”. You can find more information about the issue here: http://findbugs.sourceforge.net/bugDescriptions.html#ES_COMPARING_STRINGS_WITH_EQ
Please let me know if you have any questions.
Ayman Abdelghany.


ChristianGruen:
Thanks for the hint. We are aware of the string reference issue. In the given cases, comparisons will always succeed, as the passed on strings are used in a similar way as enums (they'll actually be one of the code strings, not only strings with same characters).

PR #1265, opened 2016-02-11
 Resources should be closed And Override Hashcode

AymanDF:
This pull request is focused on resolving occurrences of Sonar rules squid:S2095 - “ Resources should be closed”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S2095
And
squid:S1206 - “ equals(Object obj)"" and "hashCode()" should be overridden in pairs ”. You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1206
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/basex-core/src/main/java/org/basex/core/parse/StringParser.java
+++ b/basex-core/src/main/java/org/basex/core/parse/StringParser.java
@@ -48,16 +48,17 @@ final class StringParser extends CmdParser {
 
   @Override
   protected void parse(final ArrayList<Command> cmds) throws QueryException {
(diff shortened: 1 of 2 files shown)


ChristianGruen:
Thanks for the second pull request. – As the input of the Scanner class is a plain string, and as the scanner class will be discarded at the end of the function, I think it needn’t be closed (maybe I’m wrong?). Regarding the equals() function, I decided to rename it to sameAs().

PR #1273, opened 2016-03-02
Multiscreen setup

klavs:
I am experiencing problems when the main screen of GUI is located on the second screen.
The overriden `setLocation` is always limiting dialog windows to the primary screen dimensions.
I looked around the code and noticed that this method is probably defined because of some special dialog windows that get the location adjusted in relative to the main window.

I renamed `setLocation` to `setLocationWithinScreen` to explicitly mention it's function and to disable the override that was called by the superclass.

I overloaded `finish` method to clearly separate the cases that adjust location and cases that do not. Hopefully, this will make it easier for others to find the exact cases where dialog location is adjusted.

In my opinion, `setLocationWithinScreen` is still broken and should be updated or removed. I am not aware of the initial motivations to create this method (`setLocation`) so I decided to leave it here, while limiting the use of it.

### DIFF
--- a/basex-core/src/main/java/org/basex/gui/dialog/DialogAbout.java
+++ b/basex-core/src/main/java/org/basex/gui/dialog/DialogAbout.java
@@ -70,6 +70,6 @@ public final class DialogAbout extends BaseXDialog {
     p.add(newButtons(B_OK));
     add(p, BorderLayout.EAST);
 
-    finish(null);
+    finish();
   }
 }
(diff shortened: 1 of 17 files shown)


ChristianGruen:
Thanks for the pull request. After some pondering, I decided to completely remove the relative dialog adjustments, because it was only about three dialog windows (many years ago, situation was different).

I hope that the multiscreen problem is now fixed? If not, new pull requests are welcome.

PR #1304, opened 2016-06-13
PEP8 tidy up

Bengreen:
Tidy up the code (very minor) to make it PEP8 standard.

### DIFF
--- a/basex-api/src/main/python/BaseXClient.py
+++ b/basex-api/src/main/python/BaseXClient.py
@@ -19,11 +19,14 @@ Documentation: http://docs.basex.org/wiki/Clients
 
 """
 
-import hashlib, socket
+import hashlib
+import socket
 import threading
 
 # ---------------------------------
 #
+
+
 class SocketWrapper(object):
     """a wrapper to python native socket module."""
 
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Thanks; looks good.

ChristianGruen:
By the way: Tell me if you want me to announce https://pypi.python.org/pypi/BaseXClient to the mailing list!

Bengreen:
Hi Christian,
let me test it a couple of times first to see if and when it works its way through the mirrors and I can confirm finding it in search and install using pip.
Then yes would be good to announce.

KR, Ben

Bengreen:
Hi Christian,
The package has made its way through the pypi processes and is now a searchable package for python install via pip.
https://pypi.python.org/pypi/BaseXClient/8.4.4 <https://pypi.python.org/pypi/BaseXClient/8.4.4>

I will put some more effort into finishing off the .spec file for building the RPM files. It works, for BaseX, but I have not included the builtin web server which i want to create as a derived rpm as you don’t necessarily want the web server for every time you install basex.

When complete i will work out how to get this into EPEL (package release for Redhat/Centos community).

Then I might have something interesting to show you. (Integration of BaseX back into Postgres for executing queries across basex and other data sources)…. Will work out how to explain this better (effectively I can map SQL semantics onto BaseX to make basex work as part of postgres for XML data contents AND then join it to Hadoop and MongoDB data sources).

KR, Ben

ChristianGruen:
Looking forward… Thanks again for your efforts, very appreciated!

PR #1313, opened 2016-06-24
More efficient implementation of array:put(...)

LeoWoerteler:


ChristianGruen:
I expected this to come ;) …

```xquery
let $values := 1 to 1000000
return fold-left($values, array { $values }, function($array, $pos) {
    array:put($array, $pos, $array($pos) + 1)
})
```

* Before: 2.4 sec
* After: 1.1 sec

Issue #1345, opened 2016-08-25
BaseX Dockerfile: wrong permissions 


micheee:
I noticed that running basexhttp via docker leads to the following issue:

```bash
docker run basex/basexhttp
touch: cannot touch ‘/root/.m2/copy_reference_file.log’: Permission denied
Can not write to /root/.m2/copy_reference_file.log. Wrong volume permissions?
```

My suspicion is that somehow the ENTRYPOINT from the parent image (maven:3-jdk-7) is being ran on startup instead of `basexhttp`.

I am proposing a fix, but @JensErat I might need your approval ;-)

PR #1346, opened 2016-08-25
[FIX] Dockerfile: set ENTRYPOINT and CMD

micheee:
See #1345: With recent builds `basex/basexhttp` shows

    docker run basex/basexhttp
    touch: cannot touch ‘/root/.m2/copy_reference_file.log’: Permission
    denied
    Can not write to /root/.m2/copy_reference_file.log. Wrong volume
    permissions?

On startup.

This patch fixes that behaviour and _hopefully_ doesn't change anything
substantial apart from that.


micheee:
Rejected in favor of #1347

PR #1347, opened 2016-08-28
[FIX] Dockerfile: set m2 config directory

### LINKED ISSUE #1345: BaseX Dockerfile: wrong permissions

JensErat:
The upstream maven Docker image copies some files into the Maven config
directory during container startup, to fix some issues for images using
Maven at container runtime (see carlossg/docker-maven#11).

This broke startup for child images that do not run as root user (we're
adding a basex user). The same issue was described in
carlossg/docker-maven#14.

Resolves #1345, replaces #1346.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,16 +9,18 @@ RUN cd /usr/src/basex && \
     cp -r /usr/src/basex/basex-api/src/main/webapp/WEB-INF /srv
 
 # Run as non-privileged user with fixed UID
-RUN adduser --system --home /srv --disabled-password --disabled-login --uid 1984 basex && \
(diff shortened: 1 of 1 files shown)


micheee:
Thanks Jens!

PR #1359, opened 2016-09-22
proposed change for timeout in http:send-request()

cportele:
Right now, the timeout only sets the connect timeout that is active
while a connection is established, but not the read timeout while
waiting for a response.

The EXPath HTTP Client Module specification does not clearly state the
semantics of the parameter "timeout", but if the read timeout is not
considered in the implementation, a send-request() may hang forever
even though a timeout parameter has been provided.


ChristianGruen:
Makes perfect sense. Thanks for the pull request!

cportele:
Many thanks, also for the rapid response!

PR #1398, opened 2016-12-07
proposed change for supporting Jar modules packaged together with their dependencies

jonherrmann:
We have developed a Jar module for validating and comparing GML data (the XQuery Geo module did not meet all of our requirements, e.g. performance, spacial indexing, etc.). The module depends on several other libraries.

The Java UrlClassloader does not support loading Jars that are packed into other Jars ([see this paper](http://www.ibm.com/developerworks/library/j-onejar/) for more information).

This PR adds support for Jar modules that are packaged together with their dependency Jars into one single Jar. The dependencies need to be packed into the _lib_ subfolder inside the module jar. 

The _ModuleLoader_ will extract the dependencies to a subfolder __.<MODULE_NAME>Lib__ when the findClass method is called the first time and load the dependencies. There is also a unit test _installJarWithDependencies_ in _PackageAPITest_ which uses the Jar file _testJarInJar.jar_ to test this functionality.


ChristianGruen:
First impression: Looks good and elaborated! Thanks for even adding test cases. I’ll check it out in more detail soon, and I’ll probably add it to BaseX 8.6, which will be released before XMas.

jonherrmann:
Thanks for the quick reply. Please let me know if there are issues in the PR and i can help fixing it.

ChristianGruen:
Hi Jon, just some feedback in between: Sorry for the delay. The next version of BaseX will now be released in Jan 2017; be sure I’ll consider your PR just in time.

ChristianGruen:
Finally… Another response. Some minor issues I encountered:

* We still try to keep the BaseX code Java 7 compliant; could you compile the test JAR files with Java 7, or pass me on the source files?
* Nothing important, but it would be nice if you could revise the warnings output by the eclipse-cs plugin.

As soon as I can run the test, I will look at the actual code.

jonherrmann:
Hi Christian,

sure i pushed the jar compiled with JDK 1.7 to the repo. Here is also the source code:

```java
package test.jarinjar;

public class Greeter {
	public String greet(final String name) {
		return "Hello "+name;
	}
}
```

```java
package test;

import test.jarinjar.Greeter;

public class GreeterCaller {

	private final Greeter greeter = new Greeter();

	public String call(final String name) {
		return greeter.greet(name);
	}

}
```

I will check the style with eclipse again (i used IntelliJ but the cs-plugin does not seem to report all issues).

ChristianGruen:
Thanks. No worries reg. the Checkstyle stuff, it’s quickly fixed by myself.

However, the test still yields an exception (maybe due to Greeter.jar?):

```
org.basex.query.QueryException: Could not instantiate test.GreeterCaller: test/jarinjar/Greeter : Unsupported major.minor version 52.0.
	at org.basex.query.QueryError.get(QueryError.java:1367)
	at org.basex.query.util.pkg.ModuleLoader.addImport(ModuleLoader.java:132)
	at org.basex.query.QueryParser.importModule(QueryParser.java:764)
	at org.basex.query.QueryParser.importModules(QueryParser.java:750)
	at org.basex.query.QueryParser.parseMain(QueryParser.java:171)
	at org.basex.query.QueryContext.parseMain(QueryContext.java:220)
	at org.basex.query.QueryProcessor.parse(QueryProcessor.java:69)
	at org.basex.query.QueryProcessor.value(QueryProcessor.java:102)
        ...
```

jonherrmann:
The two Jars, compiled with Java 1.8 and 1.7, have identical file sizes.

The test did not work because the file sizes of the old and the new Jar were compared and finally the cached 1.8 file was used. It should work now as the timestamp is now included in the check.

ChristianGruen:
Yep! The class version error is resolved. However, the new problem is that the repo directory cannot be deleted. It could be that the JAR file, which is contained in the repo directory, is still referenced, and not unloaded after query evaluation.

There could be two reasons why you didn’t get this error:
* If you use IntelliJ, the tests may be run in a different order.
* If you don’t use Windows, the directory may be deleted, even though some files are still pinned by the file system.
```
PackageAPITest
org.basex.query.expr.PackageAPITest
installJarWithDependencies(org.basex.query.expr.PackageAPITest)
java.lang.AssertionError: Repo directory could not be deleted.
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.basex.query.expr.PackageAPITest.installJarWithDependencies(PackageAPITest.java:302)
        ...
```

jonherrmann:
The fix passed the unit tests on Windows, thanks for the hint!

ChristianGruen:
Hi Jon, finally, I made it to check your code in depth, and merge it into the master branch. Nice job! I decided to extract nested JARs at install time in order to reduce the overhead at compile time. Moreover, I further simplified the library directories by removing the Lib suffix.

I hope I didn’t break anything (because I mostly tested your own example). Feedback is welcome.

jonherrmann:
This makes sense. I tested your changes and our module is still working, so thank you very much for merging this!

ChristianGruen:
Thanks for the response. – New pull requests are always welcome!

PR #1411, opened 2017-02-10
Fix test failures

artiee:
The tests in the basex-api -project won't pass in my clean installation of Ubuntu Server -platform. This is caused due to problems in the org.basex.http.FnHttpTest and org.exquery.ns.RequestTest -classes. The problems are caused by strict comparison of headers that Http-server returns. The headers may be returned in no particular order (the actual order may depend on the platform). Also the XML attributes' order may vary. This pull request fixes these issues and addresses totally 5 failures.

Added xmlunit as dependency to help comparing parts with XML. Added new class TestUtil which is a common util for other tests.

### DIFF
--- a/basex-api/pom.xml
+++ b/basex-api/pom.xml
@@ -56,6 +56,12 @@
       <artifactId>junit</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.xmlunit</groupId>
+      <artifactId>xmlunit-core</artifactId>
+      <version>2.3.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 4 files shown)


ChristianGruen:
Thanks for the PR, very appreciated. – I wonder if we should really add another dependency just for this test. I think I’ll check if it can be solved with the existing toolset (because we can use arbitrary XQuery expressions to check test results).

ChristianGruen:
Could you please check if [my commit](https://github.com/BaseXdb/basex/commit/c0ec74f42ad45b767f9dfe982fadff322204a3c8) resolves the header issues?

artiee:
Really nice fix, yes the header issues are gone.

PR #1455, opened 2017-05-12
[MOD] Report Updated, Support SQLXML as binding, workaround for ORA

jdoe-id1:
Return affected records on an update/insert/delete statement
Add Support for SQLXML type as binding type
Add Workaround for Oracle ORA-00904 on parsing VALID update statement

### DIFF
--- a/basex-core/src/main/java/org/basex/query/func/sql/SqlExecute.java
+++ b/basex-core/src/main/java/org/basex/query/func/sql/SqlExecute.java
@@ -12,7 +12,6 @@ import org.basex.query.*;
 import org.basex.query.iter.*;
 import org.basex.query.value.item.*;
 import org.basex.query.value.node.*;
-import org.basex.query.value.seq.*;
 import org.basex.util.*;
 
 /**
@@ -39,7 +38,7 @@ public class SqlExecute extends SqlFn {
     if(!(obj instanceof Connection)) throw BXSQ_CONN_X.get(info, id);
     try {
       final Statement stmt = ((Connection) obj).createStatement();
-      return stmt.execute(query) ? iter(stmt, true) : Empty.ITER;
+      return stmt.execute(query) ? iter(stmt, true) : Int.get(stmt.getUpdateCount()).iter();
     } catch(final SQLException ex) {
       throw BXSQ_ERROR_X.get(info, ex);
     }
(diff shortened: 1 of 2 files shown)


basex-core/src/main/java/org/basex/query/func/sql/SqlExecute.java
```diff
@@ -39,7 +38,7 @@ public Iter iter(final QueryContext qc) throws QueryException {
     if(!(obj instanceof Connection)) throw BXSQ_CONN_X.get(info, id);
     try {
       final Statement stmt = ((Connection) obj).createStatement();
-      return stmt.execute(query) ? iter(stmt, true) : Empty.ITER;
+      return stmt.execute(query) ? iter(stmt, true) : Int.get(stmt.getUpdateCount()).iter();
```
ChristianGruen: Thanks for your pull request. – I didn’t check the code in detail yet, but isn’t it wrong to return an integer if the SQL statement is non-updating?

jdoe-id1: Sorry, you absolutely right! I have to check the return value of getUpdateCount as documented. And in case of an empty result of a select statement an empty sequence must returned as before!

jdoe-id1: Ok, I did some further investigation and tests on oracle, mysql and postgresql.

If a SELECT statement was processed by execute, it will be always return true, even no result was returned ("select * from sometable where 1=0"). So an empty sequence will return within xquery anyway and not change in current behaviour of BaseX. 

If it was an update (insert, delete or ddl) statement, execute returns always false and the number of affected record can received by getUpdateCount. getUpdateCount will return 0 if no rows are affected ("update someTable set someColumn='x' where 1=0"). 

In consequence it will not mixed up with empty sequence as result of an empty select.

So the patch should be fine as it is :)

ChristianGruen: Sounds completely reasonable! Thanks for trying 3 different RDBMS. I’ll additionally check SQLite, and I’ll merge your PR if everything works as expected.

ChristianGruen:
A great PR, thanks. I have changed some minor issues (), and I have updated our documentation (http://docs.basex.org/wiki/SQL_Module#sql:execute-prepared). A new stable snapshot is available (http://files.basex.org/releases/latest/), BaseX 8.6.4 is scheduled for end of this week.

PR #1461, opened 2017-06-05
Removed outdated Checkstyle module

infstylecheck:
Running `mvn checkstyle:checkstyle` currently crashes because the configuration files have the module `StrictDuplicateCode` which is outdated. This PR removes the outdated module from the configuration files so that `mvn checkstyle:checkstyle` now does not crash.


ChristianGruen:
Thanks, looks good.

PR #1465, opened 2017-06-22
Completed Chinese translation

donaldjohn:
Hi,
  I added Chinese translation to BaseX, please accept my pull request, so that BaseX can be easier for Chinese users.


ChristianGruen:
@donaldjohn, thanks for the translation, very appreciated!

PR #1468, opened 2017-06-30
Changed java.util.zip to zip4j in DirParser.java to keep compatible with some weird zip files

donaldjohn:
I came across some weird zip files, those files be uncompressed with winrar in windows and unzip in linux, but can't parsed in baseX or some other uncompress tools. the baseX throws Exception with message "invalid entry compressed size (expected 4294967295 but got 13733720 bytes)". I debuged BaseX core and located the code that throws this exception. It throwed by java.util.zip. 
Firstly, I changed java.util.zip to commons-compress, it stopped throwing exception but commons-compress can only find 1 entry in my zip file(actually there were 177 entries).
Then I turned to zip4j, it works. BaseX can parse my zip files correctly now.
Perhaps zip4j can tolerate more fault than java.util.zip.


ChristianGruen:
Dear Donald, thanks for the proposal. I tend to stick with the default Java ZIP streams, because it seems to work fine if the input is valid. Maybe you have any chance to unzip the files before passing them on to BaseX?

For further pull requests, it would be great if you could try to keep the indentation of the original files (such that only those lines in the code will be replaced that have actually been rewritten by you).

apb2006:
Hi Donald,
I think the extra weight of `zip4j`, both as another dependancy  and size of 120+kb is difficult to justify just to cope with a small number of possibly dubiously formated files. On the other hand I feel the default Java ZIP streams have poor write performance and have seen reports that `zip4j` performs better. I have not done any tests to confirm this. 

A solution here could be to create an extension see http://docs.basex.org/wiki/Repository#Java_2 to wrap the zip4j library. Installable using the REPO commands. This could be installed/called by developers where it provided advantages. 

I have created a number of similar packages wrapping Java libraries at https://github.com/expkg-zone58 . Maybe you could create something similar? Indeed, I would be happy to add you to expkg-zone58 if you would like to host it there.

/Andy

donaldjohn:
@ChristianGruen I am sorry about the indentation, the file was formatted by IDEA automatically. I will pay attention to it next time.
Since zip4j is not fully tested, it is wise to keep java.util.zip

donaldjohn:
@apb2006 Thanks a lot, your suggest is very good, it is really what I wanted that extending the baseX without change it's code.  
you can add me to expkg-zone58, I will figure out how to implement the extension some time later. to be honest I am new to baseX, it may take me a little time to dig deep into it.

apb2006:
@donaldjohn Great. I have sent an invitation to start the process.  Dont worry about the time, any questions email me or create an issue at https://github.com/expkg-zone58/expkg-zone58.github.io/issues

PR #1491, opened 2017-08-21
Fix compilation with Java 9

FauxFaux:
In Java 9, to be released in 30 days, some of basex ceases to compile or work:

 * `java.lang.Module` has been added, causing star imports of classes named `Module` to fail: fixed by adding explicit imports.
 * swing's `BasicComboPopup` constructor type has been changed. This is probably a JDK bug, but it's probably a bit late to fix it.
 * An unchecked cast has become invalid, replaced with wildcards.
 * A single test depends on some JAXB implementation, which has been removed from the JDK in 9. I've added one from Maven Central, at `test` scope. The state of JAXB implementations in Maven is a bit of a disaster. I doubt the test cares about a specific implementation; it's a very small use. Maybe there are some parts of the application which need these jars not a test scope? I don't know how you want to handle this.

All the tests pass, except ProfModuleTest#human, which was broken anyway. (170a3e2eaf3b099304d3cff951fb53e65e5192eb?)

### DIFF
--- a/basex-core/pom.xml
+++ b/basex-core/pom.xml
@@ -19,6 +19,22 @@
       <artifactId>junit</artifactId>
     </dependency>
     <dependency>
+      <groupId>javax.activation</groupId>
+      <artifactId>activation</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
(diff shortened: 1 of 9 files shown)


ChristianGruen:
Great PR, thanks Chris! I have just fixed the broken test, and I will remove the single test (`SAXSerializerTest` I think it is) that relies on JAXB.

PR #1497, opened 2017-09-18
[FIX] sqlExecutePrepared. Disable all oracle exceptions

jdoe-id1:
An other oracle exception for a valid statement was thrown. So we changed it to ignore any oracle exception of getParameterMetaData.


ChristianGruen:
Thanks

jdoe-id1:
Thx too :)

PR #1503, opened 2017-09-28
File open dialog opening in background (at least on Mac OS)

dirkk:
Hi Christian,

I noticed a minor bug in the GUI on this Mac OS I now use (and still loath...). When creating a new database and opening an input xml it would open the file chooser dialog in the background of the "New Database" dialog.

Then I remembered that I now write Java code for a living and though I could fix it myself ;). You see the result here. I can't remember seeing this bug on my lovely Linux OS, so I assume it might only materialize on Mac OS. However, I guess @micheee should be able to confirm/deny. But looking at the code it actually does make sense, because the parent window of the file chooser was always the main BaseX GUI. I would have liked to completely replace the `gui`variable, but there is this (weird, historic?) "Simple File Chooser" global option, which is needed. So I instead opted for another `parent`variable.

PS: Maven is still coming, but I am in a train with unstable internet (Schwarzwald!), so I couldn't work on it right now. Maybe in the next couple of days, but I am unfortunately still swamped.

### DIFF
--- a/basex-core/src/main/java/org/basex/gui/dialog/DialogExport.java
+++ b/basex-core/src/main/java/org/basex/gui/dialog/DialogExport.java
@@ -72,7 +72,7 @@ public final class DialogExport extends BaseXDialog {
    * Opens a file dialog to choose an XML document or directory.
    */
   private void choose() {
-    final IOFile io = new BaseXFileChooser(CHOOSE_DIR, path.getText(), gui).select(Mode.DOPEN);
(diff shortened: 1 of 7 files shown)


ChristianGruen:
Dirk, thanks for the PR! I’ll check if we can get rid of the old legacy file chooser (should be about time).

ChristianGruen:
Hi Dirk, I’ve dropped the old file chooser, and I have revised the window reference handling, inspired by your pull request. A new snapshot is available, I hope it resolves the problems. If not, feel free to reopen this PR or another issue. – Greetings to whereever you left the Schwarzwald train.

PR #1489, opened 2017-08-18
Maven Central

dirkk:
As discussed the little update to the pom.xml required to comply the OSSRH standards.

You will have to sign the packages with pgp, for setting it up see http://central.sonatype.org/pages/working-with-pgp-signatures.html for instructions. Then you will have to modify your settings.xml from Maven (.m2/settings.xml) like this:

```
<settings>
  <profiles>
    <profile>
      <id>ossrh</id>
      <activation>
        <activeByDefault>true</activeByDefault>
      </activation>
      <properties>
        <gpg.executable>gpg2</gpg.executable>
        <gpg.passphrase>the_pass_phrase</gpg.passphrase>
      </properties>
    </profile>
  </profiles>
</settings>
```

The actually signing will be done by the `maven-gpg-plugin`. 

When this is set up it should automatically release the artifacts to OSSRH when you run `mvn deploy`.

However, there is still one issue left: I asked in the jira repository that you are being added to the project (https://issues.sonatype.org/browse/OSSRH-33622?focusedCommentId=428864&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-428864). I am not sure whether this is the correct approach, but it looks like this. Right now I would expect you do not have the required permission to upload a new release.

PR #1513, opened 2017-11-01
[MOD] releasing non-snapshots to maven central works for me now

dirkk:
Ok, so I finally managed to find time for a follow-up on https://github.com/BaseXdb/basex/pull/1489 .... I guess thanks Martin Luther for the additional public holiday :+1: 

So, I basically didn't change anything: The part with `doclint-java8-disable` is only because I use java8 and the JavaDoc is incorrect (see https://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete for more info. Of course it would be better to fix the actual issue...)

The other plugin just moved to the `release` profile.

However, I now tried this: So running 

    mvn clean deploy -DskipTests -P release

deploys the stuff to osshr. But after that it is still not deployed: For this you have to login in the Nexus and search here https://oss.sonatype.org/#stagingRepositories for basex. When the deploy was successful you should see a new repository, something in the format `orgbasex-1010` (artifactId without the dots and a number starting at 1000). In the web interface you can then release it with the "Close" Button (who named this???).

Of course, this can be automated. But I am unsure what you currently use and I also don't want to accidentally release some crap... If you use the release plugin and do something like 

    mvn release:clean release:prepare
    mvn release:perform

Otherwise, if you don't use it and stick with the deploy phase after runnign deploy the following

    mvn nexus-staging:release

should do the same as the web interface "Close" button.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,6 +323,40 @@
               <appendAssemblyId>false</appendAssemblyId>
             </configuration>
           </plugin>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
+            <version>1.6</version>
+            <executions>
+              <execution>
+                <id>sign-artifacts</id>
+                <phase>verify</phase>
+                <goals>
+                  <goal>sign</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
+      <id>doclint-java8-disable</id>
+      <activation>
+        <jdk>[1.8,)</jdk>
+      </activation>
+
+      <build>
+        <plugins>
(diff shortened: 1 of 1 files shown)


dirkk:
Also, btw you can add to your personal settings.xml also your passphrase, i.e. something like this:

```
<settings>
  <profiles>
    <profile>
      <id>ossrh</id>
      <activation>
        <activeByDefault>true</activeByDefault>
      </activation>
      <properties>
        <gpg.executable>gpg2</gpg.executable>
        <gpg.passphrase>PASSPHRASE</gpg.passphrase>
      </properties>
    </profile>
  </profiles>
  <servers>
    <server>
      <id>ossrh</id>
      <username>itismetravelling</username>
      <password>PASSWORD</password>
    </server>
  </servers>
</settings>
```

PR #1522, opened 2017-11-21
[ADD] options to sql:execute and execute-prepared to set query-timeout

jdoe-id1:
Hi,
add an extension to sql:execute and sql:execute-prepared. New optional options allow to specify a query-timeout. Helps to avoid longtime running query on server while xquery script has already terminated.

usage: sql:execute($conn, 'select max(col) from very_big_table', , map {"query-timeout": 10 })

Timeout is given in seconds. see: https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setQueryTimeout-int-

I do not provide a new xquery function like sql:set-query-timeout($preparedStatement, 10) because sql:execute will not benefit.

Cheers
Jan


ChristianGruen:
Looks good! To be sure: Can you confirm that an [SQLTimeoutException](https://docs.oracle.com/javase/7/docs/api/java/sql/SQLTimeoutException.html) will be raised if the timeout is exceeded? If yes, I might add an additional error code. And I guess I’ll rename the option to `timeout`. Both changes would be in analogy with [xquery:eval](http://docs.basex.org/wiki/XQuery_Module#xquery:eval).

jdoe-id1:
Yes It will raise SQLTimeoutException: "java.sql.SQLTimeoutException: ORA-01013: Benutzer hat Abbruch des aktuellen Vorgangs angefordert"

So I renamed option into "timeout" and add extra exception to basex.

jdoe-id1:
If you satisfied with the change, I will create a new pull request and merge all changes into a single commit.

ChristianGruen:
Excellent! I’ll merge your PR later this day (GitHub will take care of squashing your commits into a single one).

jdoe-id1:
Thanks!

ChristianGruen:
Merged. I’ve additionally doneo some minor changes (checkstyle issues; allowing assignment of negative values, because the JDBC driver should raise an error here, that will eventually be caught).

ChristianGruen:
Updated documentation: http://docs.basex.org/wiki/XQuery_Module#sql:execute

jdoe-id1:
Great! 
Beside, the new timeout error is missing in the Wiki!?

ChristianGruen:
Thanks for the observation. I noticed I’ve actually messed up the XQuery Module instead of enhancing the SQL Module. Should be fixed again: http://docs.basex.org/wiki/SQL_Module

PR #1524, opened 2017-11-27
Created package for Python client API.  Modified examples to use it.

jonathanrobie:


basex-api/src/main/python/QueryExample.py
```diff
@@ -6,7 +6,7 @@
 #
 # (C) BaseX Team 2005-12, BSD License
 
-import BaseXClient
+from BaseXClient import BaseXClient
```
jtauber: trailing whitespace

PR #1526, opened 2017-11-28
[MOD] add clob handling

jdoe-id1:
Hi,

in case of an Oracle CLOB column only a "pointer"-value will returned by toString. 
So here some special handling for clobs is need. Beside Clob.getSubString also RecordSet.getString() will work in this Situation. But we followed the way as xmltype was handled here.

BTW: In case of execute-prepared big clob works fine, because the PreparedStatement.setString the counterpart of RecordSet.getString is used.

Best Regards

Jan


ChristianGruen:
I couldn’t test it, but I assume it should be fine. Have you tried it with any other relational database as well?

jdoe-id1:
As far as I know beside Oracle only DB2 support CLOBs. We test it only on Oracle 11/12.
Other Database should not affected.
Thanks for pull :)

PR #1531, opened 2017-12-11
updated C API for newer authentication method

smallfriex:
updated C API, so it works with v8.0+ of basex
this was tested with the example code and worked
should also work with older versions
and I have extern "C"'d the headers so this API
can be compiled with a c or c++ compiler


ChristianGruen:
Looks good! I have merged your PR, thanks again.

PR #1533, opened 2017-12-18
Ignorehostnameverification

dcore94:
as discussed in email exchange a patch for separating lax hostname verification from igrnoring certs altogether in HTTPS.


ChristianGruen:
Looks fine! All I did was choosing an even shorter name for the option (mostly because all other existing options have shorter names).

PR #1538, opened 2018-02-08
Authentification adapted to BaseX 8

thormu:
I have modified the authentification to connect to BaseX 8 server.

### DIFF
--- a/basex-api/src/main/c/basexdbc.c
+++ b/basex-api/src/main/c/basexdbc.c
@@ -1,6 +1,5 @@
 /**
  * basexdbc.c : communicate with BaseX database server
- * Works with BaseX 7.x (but not with BaseX 8.0 and later)
  *
  * Copyright (c) 2005-12, Alexander Holupirek <alex@holupirek.de>, BSD license
  */
@@ -111,7 +110,10 @@ basex_authenticate(int sfd, const char *user, const char *passwd)
 	char ts[BUFSIZ]; /* timestamp returned by basex. */
 	char *md5_pwd;   /* md5'ed passwd */
 	int ts_len, rc, i;
-
+    char* item;
+    char* item1;
+    char* item2;
+    
 	/* Right after the first connect BaseX returns a nul-terminated
          * timestamp string. */
 	memset(ts, 0, BUFSIZ);
@@ -120,17 +122,17 @@ basex_authenticate(int sfd, const char *user, const char *passwd)
 		warnx("Reading timestamp failed.");
 		return -1;
 	}
-	ts_len = strlen(ts);
 
+    item = ts;
+	item1 = strtok(item, ":");
+    item2 = strtok(NULL, ":");
+    
 #if DEBUG
 	warnx("timestamp       : %s (%d)", ts, strlen(ts));
 #endif
 
 	/* BaseX Server expects an authentification sequence:
-           {username}\0{md5(md5(user:realm:password) + timestamp)}\0 */
- 	/* legacy - 
-        /* {username}\0{md5(md5(password) + timestamp)}\0 */
-        
+ 	 * {username}\0{md5(md5(password) + timestamp)}\0 */
 
 	/* Send {username}\0 */
 	int user_len = strlen(user) + 1;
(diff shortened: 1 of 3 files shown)


ChristianGruen:
Thanks for your pull request.

Just recently, we got and merged another pull request for the C client (https://github.com/BaseXdb/basex/pull/1531, by @smallfriex), which provided support for both older versions and BaseX 8. It seems it isn’t working for you? Did you try to find out what might go wrong in the 8.0 branch of the current code?

thormu:
Thanks you for your reply.
I have passed a modified argument to the md5 calculation. By comparing the authentification with the C#-BaseXClient, it gives me an idea how to modify it. So for me it works now. If you have another working solution , you can add it to your existing c client.

smallfriex:
Hi Thormu. I've been using my code for the past few weeks, and it works well. This is C code, not C#, but should give you an idea of how to make the C# interface work with both newer and older BaseX clients, if you haven't done that already.

Commit: 8a82c71b1a08bea964cffd6e74bb61e67c599faf. The example.c shows how to use the api code.

thormu:
No, the C# Client is working well, we use it to. The old C Client had a problem with the authentification.
we have modify the md5 calculation call in the file basexdbc.c . The comment in this file gives a hint, that the last version works only with BaseX 7, but we use BaseX 8.6.

smallfriex:
Oh, okay. I thought I'd fixed that. Sorry to hear it doesn't work for you. I'm using 8.5.1 of BaseX. I should've updated the comment - whoops. I'll take a look at the C# code now and see what differences there may be.

One thing, when I ran valgrind on my code, it did complain about a non-initialised variable in the md5.c code, but since my code was working, I didn't investigate that further.

smallfriex:
latest C# and C code appear to logically do the same thing and send the same digests, however the C# code is much clearer and easier to understand.

I can only assume you didn't have the latest code when you experienced authentication problems. Unless something is working different on Windows versus linux (where I tested my changes). Either way, it would be interesting to see your changes to the basedbc.c file.

thormu:
You can check the basedbc.c file from line 145 to 155. There i build up a string to pass it later to the md5 calculation.

ChristianGruen:
Hi @thormu, I think we should stick with the client version from @smallfriex. If you still believe it’s doing something wrong, could you please add your changes on top of his code base (which supports older versions of BaseX as well)? Thanks in advance.

moniwe:
i tested your version in gitHub on windows, and the authentification fails for BaseX 8.6.7, whereas thormus version works.

thormu:
I think, the problem is the link under http://docs.basex.org/wiki/Clients for the C-Client. If you follow this link, you don't see your modified code  for BaseX 8.6. So i have used this old code and adapted it.

ChristianGruen:
> I think, the problem is the link under http://docs.basex.org/wiki/Clients for the C-Client. If you follow this link, you don't see your modified code for BaseX 8.6. So i have used this old code and adapted it.

@thormu: Good to know; I have updated the link.

> i tested your version in gitHub on windows, and the authentification fails for BaseX 8.6.7, whereas thormus version works.

@moniwe: Did you test the version that was reference in the BaseX wiki, or the version that was discussed here (i.e. the one that is now referenced in our documentation)?

moniwe:
i checked out https://github.com/BaseXdb/basex.git and tested the c client in the \basex-api, since i thought that the commit "updated C API for newer authentication method" was meant to solve the authentification.

smallfriex:
Looking through the code, my change and Thormu's code do the exact same thing (and both work). Even my comments have been reversed, which suggests it was branched from code before my commit.

My vote would be to keep my change, and I'll update the comment header later today. 

Another observation is that, as it stands, basexdbc.c won't compile on windows anyway - it's got references to err.h, netdb.h etc - i.e, nix headers, so I'm guessing that Thormu has used some functions from basexdb.c and put them in another harness?

So maybe what we should do, is create another file, called basexdbc_win.c, with Thormu's code that would compile for windows. We could even make an example_win.c, when someone has the time to spare.

At the moment, the c++ interface uses boost, so that is platform independent. Of course, not everyone wants to use boost. And some embedded platforms will support c but not c++, but I think Windows will always have c++ compilers. So, there's probably a work-around for most people right now.

thormu:
Yes, you can keep your changes and add it to the master branch. I would like to have an windows version too, because maybe someone prefers the C-Client version.
Thanks for spend time to fix it.

ChristianGruen:
I have just merged a little pull request from @smallfriex, and I closed this pull request. Pull requests for a Windows version are welcome (it would be great if they were based on the existing C code).

PR #1551, opened 2018-03-20
improve Unix startup scripts to use exec, not fork

bubak4:
- minor fix to Unix provided startup/shutdown scripts
- previous versions just called another process (fork)
- current version replaces the running process (exec)
- advantage: for various process supervisors it is better to use exec, not
  fork, since the shell script PID becomes the real process PID
- new (current) script can be used with
  supervisor (http://supervisord.org/) without any change, exec-ing the
  process means the supervisor can correctly track the started server

### DIFF
--- a/basex-api/etc/basexhttp
+++ b/basex-api/etc/basexhttp
@@ -17,4 +17,4 @@ CP=$MAIN/target/classes:$MAIN/lib/*:$CORE/lib/*:$MAIN/lib/custom/*:$CLASSPATH
 BASEX_JVM="-Xmx2g $BASEX_JVM"
 
 # Run code
-java -cp "$CP" $BASEX_JVM org.basex.BaseXHTTP "$@"
+exec java -cp "$CP" $BASEX_JVM org.basex.BaseXHTTP "$@"
--- a/basex-api/etc/basexhttpstop
+++ b/basex-api/etc/basexhttpstop
@@ -14,4 +14,4 @@ CORE="$( cd -P "$MAIN/../basex-core" && pwd )"
 CP=$MAIN/target/classes:$MAIN/lib/*:$CORE/lib/*:$MAIN/lib/custom/*:$CLASSPATH
 
 # Run code
-java -cp "$CP" $BASEX_JVM org.basex.BaseXHTTP "$@" stop
+exec java -cp "$CP" $BASEX_JVM org.basex.BaseXHTTP "$@" stop


holu:
I would also opt for changing `bin/basexgui`:

```diff
# Run code
-java -cp "$CP" $BASEX_JVM org.basex.BaseXGUI "$@"
+exec java -cp "$CP" $BASEX_JVM org.basex.BaseXGUI "$@" &
```

- This would allow the GUI to be started from the shell without being a child of it (and closed, if the shell is closed.)

bubak4:
I do not think it is a good idea -- I believe in Unix you usually suppose the process started from a terminal will occupy your stdin and stdout (ie. all the shell text utils like cat, grep etc.) and you have always the option to add '&' to start the process in the background, if you want to.

Forcing the process to start in the background like you propose, you make a decision for the user which is not desirable imho.

Adding an exec in this scenario is also counter-intuitive, since it is custom to use fork, when executing from terminal (for example https://www.tldp.org/LDP/Bash-Beginners-Guide/html/sect_01_03.html).

ChristianGruen:
Thanks for the suggestions! So we should probably update `basexserver` as well?

holu:
Thanks for feedback and edit regarding `exec`.  I agree, that `exec` does not break anything, but is not necessary in this case either.

>  you have always the option to add '&' to start the process in the background, if you want to.

> Forcing the process to start in the background like you propose, you make a decision for the user which is not desirable imho.

I might be wrong, but doesn't it make a difference where the '&' is placed and what the effect is?

The invocation of the current startup script with '&' is putting the bash script to background, but leaves it and the (gui) java process connected to the terminal session. If the terminal session ends, basexgui is SIGHUPed (if its not `disown(1)`ed before), right?

If the '&' is placed within the script the `java` process is put to background, decoupled from the terminal session and continues to run even if the terminal session is terminated?

To me, the later is more what a user would expect. `basexgui` is a graphical app (more like a browser than a shell text utils like `cat(1)`, `grep(1)`). When started, by chance, from the shell (and not from a dock or another application launcher) the gui should continue to run even if the startup terminal is closed.

(I don't want do hijack your PR. And, since it is not related to exec anymore, please let me know if I should open a separate PR.)

bubak4:
You are right and you will get the behaviour described if you put an '&' into the script itself. Your process will outlive the parent and will end up attached to your init/systemd daemon. I also agree that basexgui should be ragarded as a browser more than cat.

I think this should be a separate PR.

Only for consideration:

- your behaviour is maybe 'too clever' and could easily outsmart a user -- I would not expect such a behaviour on a program I execute
- for example chromium started from terminal stays attached to the terminal session as well as other GUI programs I tend to run
- by _not_ putting the '&' into the script you still give user a chance to easily create custom script with behaviour you described
- by putting it there means, you decide for the user and he has no chance to override enforced behaviour unless editting your source code
- maybe it would be possible to provide XDG .desktop launcher and integrate basex GUI seamlessly to desktop

holu:
Hi @bubak4, thanks for the discussion and your considerations!

ChristianGruen:
@bubak4: No problem; all your comments are appreciated.

PR #1565, opened 2018-05-04
changes for latest version of openssl

smallfriex:
Changes to work with type changes in newer version of openssl v1.1.0
- also compatible with older openssl v1.0.1
- this has been tested and works
- this newer version of openssl/crytpo library is installed on latest Ubuntu 18.04 LTS


ChristianGruen:
This one has slipped through my fingers... Thanks for the pull request, looks good.

PR #1605, opened 2018-08-06
Websockets

jfinckh:
* Add HttpSession to the websockets
* Remove get/set/delete -> use Session/s-module for attributes


basex-api/src/main/java/org/basex/modules/ASession.java
```diff
@@ -42,16 +42,15 @@
       final Object ws = qc.getProperty(HTTPText.WEBSOCKET);
       if(ws != null) {
         session = ((WsAdapter) ws).httpsession;
-        if(session != null) return;
       } else {
         final Object req = qc.getProperty(HTTPText.REQUEST);
         if(req == null) throw BASEX_HTTP.get(null);
```
ChristianGruen: Another minor one (sorry for being picky ;-)… If no session is available, BASEX_HTTP should be raised.

Issue #1620, opened 2018-09-15
sessions:set(xs:string, xs:string, item()) cannot be called with (xs:string, xs:string, item()+).


jfinckh:
sessions:set has an different behavior than session:set.
While session:set can work with sequences, sessions:set cant.

The documentation specifices for both functions: 

> $value as item()*

But session:set takes an Value.

**Code:**
```sessions:set($id,$chat:messages,($msg,$message))```
**Error:**
```Function set(xs:string, xs:string, item()) cannot be called with (xs:string, xs:string, item()+).```

PR #1621, opened 2018-09-15
Fix sessions:set

jfinckh:
https://github.com/BaseXdb/basex/issues/1620


ChristianGruen:
Thanks for the immediate fix!

jfinckh:
Thanks for the immediate merge :D

PR #1629, opened 2018-10-01
[ADD] add check for origin

jfinckh:
Close connection if request from the wrong origin


ChristianGruen:
I’d tend to close this one. We should rather look for a solution that works for all of our web services (or let proxies – or the web server? – do the job).

PR #1635, opened 2018-10-30
macOS: Adjust apple system property

holu:
Move Apple specific system properties to OSX classes.
Set useScreenMenuBar property for all JDKs.


ChristianGruen:
Looks good!

PR #1668, opened 2019-02-28
Corrected wadl:resources/@base in RestXqWadl

vcestone:
In the wadl generated by the `rest:wadl()` function, we noticed that the `wadl:resources/@base` contains redundant prefix that clashed with the `wadl:resource/@path`. This issue emerged while importing the resulting wadl into tools such as Postman. The proposed pull request should fix it in all cases, including deployment in application containers such as Tomcat (where there is a context path too).


ChristianGruen:
Looks good!

PR #1844, opened 2020-04-23
Complete Junit5 migration

dimitarp:
* fixed imports
* fixed changed assertEquals signature (in JUnit5 the message is the last argument)
* refactored parameterized tests (JUnit4 Parameterized-Rule is no longer available)
* added missing JUnit dependencies and set a newer version of Junit5
* added missing volatile in the Job class
* some other improvements in test code
:)


ChristianGruen:
Awesome… Thanks a lot for the effort, Dimitar!

basex-core/src/test/java/org/basex/server/SessionTest.java
```diff
@@ -545,27 +539,25 @@ public void queryBind2() throws IOException {
...
+  public void queryError2() {
+    assertThrows(BaseXException.class, () -> session.query("(1,'a')[. eq 1]").execute());
   }
 
-  /** Runs an erroneous query.
-   * @throws IOException expected exception*/
-  @Test(expected = BaseXException.class)
+  /** Runs an erroneous query. */
+  @Test
+  // FIXME: it seems that an error is thrown on the first next() call even though it should be ok
```
ChristianGruen, line 556: @dimitarp The test doesn’t trigger an error on my system. How can this be reproduced?

ChristianGruen: …and I was too quick. I’ll have a look into it.

ChristianGruen: Fixed: The existing test was deficient, as it was the first `next()` call that raised the error.

This was different in (much) earlier versions of BaseX, in which the query API didn’t cache all results before delivering them to the caller.

dimitarp: yes, I guessed so, but I thought you could know better what is the cause ;)

PR #1854, opened 2020-05-07
Minimize Reflection + Jing Update

dimitarp:
### Minimize Reflection
When defining a function, pass a `java.util.function.Supplier` instead of the class which implements the XQuery function. In terms of syntax nothing is lost: instead of passing `MyFunction.class` one uses the default constructor reference `MyFunction::new`. In terms of performance, there should be a small gain since we don't use reflection any more.
### Jing Update
The maven groupId of the Jing library has changed to `org.relaxng`. I added some tests to verify there are no (big?) regressions using the new version.

### DIFF
--- a/basex-api/src/main/java/org/basex/query/func/ApiFunction.java
+++ b/basex-api/src/main/java/org/basex/query/func/ApiFunction.java
@@ -5,6 +5,7 @@ import static org.basex.query.util.Flag.*;
 import static org.basex.query.value.type.SeqType.*;
 
 import java.util.*;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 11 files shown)


ChristianGruen:
Thanks again! Will check this later today (and maybe split it into two separate commits).

ChristianGruen:
I have updated the Jing artifacts… and I have dropped the tests. The reason is that I try to keep all tests runnable without external libraries. Maybe I should reconsider that (or we could include checks in the XQuery tests, and run them only if RelaxNG or other libraries are available in the classpath).

PR #1949, opened 2020-10-18
Refactor PHP client

igneus:
Code of the PHP client has to be modified for use in modern PHP projects, so I tried to fix this.

* directory structure adjusted with [PSR-4 autoloading](https://www.php-fig.org/psr/psr-4/) in mind, namespace added
  -- i.e. the code as is does not use autoloading, but the `basex-api/src/main/php/BaseXClient` can be copy-pasted to a project with PSR-4 autoloading enabled and with a small bit of configuration in `composer.json` will work fine

Then I added some more improvements and updated some of the most archaic code features, while trying not to break compatibility with "reasonably old" outdated PHP versions.

* code style according to the [PSR-2](https://www.php-fig.org/psr/psr-2/) community standard enforced by [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)
* various code smells fixed (see commit messages)
* `Query` implements the `Iterator` interface

### DIFF
--- a/basex-api/src/main/php/AddExample.php
+++ b/basex-api/src/main/php/AddExample.php
@@ -6,7 +6,9 @@
  *
  * (C) BaseX Team 2005-12, BSD License
  */
-include("BaseXClient.php");
+include_once 'load.php';
+
+use BaseXClient\Session;
 
 try {
     // create session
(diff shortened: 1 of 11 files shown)


ChristianGruen:
Thanks for the pull request. I’ll merge it without further revisions (it’s a clear improvement over the existing client binding).

PR #2030, opened 2021-08-03
Signature resilient to missing query

dcore94:
We've found out (but not clearly identified) that with some combinations of browsers/proxies the body is not an empty string "" but rather an empty sequence. This is causing error when opening the database view of the dba. ("_Request body must be xs:string, supplied ()_")
We propose to make the query parameter optional.

### DIFF
--- a/basex-api/src/main/webapp/dba/databases/resources/db-query.xqm
+++ b/basex-api/src/main/webapp/dba/databases/resources/db-query.xqm
@@ -24,7 +24,7 @@ declare
 function dba:db-query(
   $name      as xs:string,
   $resource  as xs:string,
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Hi Marco, looks like a very non-intrusive and reasonable change.

Could you check for us if the same error can be observed when an empty query is saved in the Query panel (i.e., when `dba:query-save` is called)?

dcore94:
Hi Christian, yes. we are able to confirm that the issue arises also in that case.

ChristianGruen:
Perfect; I’ll merge your PR and add another question mark for saving queries.

PR #2043, opened 2021-10-09
Add raku client

wayland:
Hi!  

In these commits, I've added a raku client.  I started with the Perl client, and ported it across fairly straight.  Then I:
- Separated all the protocol information out into a separate JSON file
- Made the client use the protocol information to figure out parameters, etc
- Made the client accept named parameters (mostly as well, but sometimes instead)
- Replaced a bunch of almost identical function calls with a FALLBACK
- Added support for the UPDATING command by modifying the JSON file

Note that all of the above (including depending on the JSON file) is entirely possible in Perl too.  

HTH,


wayland:
My bad; you can't mix named and positional parameters in Perl.

PR #2047, opened 2021-11-05
Gzip handler - decompress requests

jb8748:


jb8748:
GZIP: true must be set in options for any GZIP compression/decompression to happen

Aim: allow RESTXQ calls where Content-Encoding is set to gzip.

Currently when I make a compressed POST request Jetty does not inflate the request. Responses are compressed as expected.

The inflate buffer size is not explicitly set by the code. And appears to default to -1.
A value greater than 0 is required for inflation to happen in GZIPHandler.
This pull request is to set that value explicitly.

ChristianGruen:
Just fine, thanks

PR #2049, opened 2021-11-26
Fix the Bindings dialog tab order

dliakh:
Change tab traversal order of the fields on the Bindings form so the text edit fields are traversed in rows instead of columns


dliakh:
Please see the 'tracker' issue (opened for this specific case/change)
https://github.com/BaseXdb/basex/issues/2050

ChristianGruen:
Merged, after rewriting the code (setNextFocusableComponent is deprecated).
Thanks, though.

PR #2110, opened 2022-05-21
fix in EQNAME regexp

hfmanson:
In EQNAME the closing brace also needs to be escaped.
Found out while using BaseX in android which gave a Pattern syntax error


ChristianGruen:
Interesting to learn about the differences with Android. Thanks.

PR #2149, opened 2022-11-09
[FIX] macOS Perform Quit

micheee:
* [ ] Canceling the close when the editor has been changed


micheee:
What's not yet working: 

* Canceling the close when the editor has been changed.

PR #2154, opened 2022-12-28
Allow running BaseX from a fat jar

GuntherRademacher:
While trying to use BaseX in an application that is packaged as a fat jar, I found a few problems:

- Prop.applicationDir assumes that it is passed a file URL, which in this setup does not hold for Prop.LOCATION (which is something like "rsrc:basex-10.5-SNAPSHOT.jar"). This causes a FileSystemNotFoundException stack to be logged
- ModuleLoader.close closes the classloader when it is an URLClassLoader. However in this setup, the default classloader is an URLClassLoader and it must not be closed
- an XQuery module URL, in this setup, might be something like "jar:rsrc:xquery-module-test-plain.jar!/m0.xqm". This however not accepted by IOURL.isValid, because it insists on a '/' character following the protocol
- IOURL.inputStream use an HTTP connection, which does not work for jar URLs. For those it is sufficient to use URL.openStream
- when resolving a module URI containing a parent step, e.g. "../m4/m4.xqm", the parent step must be resolved as well. This can be accomplished by using the URL(URL, String) constructor

I have created a test that demonstrates all of this here: https://github.com/GuntherRademacher/xquery-module-test

This PR contains fixes to make it work.

### DIFF
--- a/basex-core/src/main/java/org/basex/io/IO.java
+++ b/basex-core/src/main/java/org/basex/io/IO.java
@@ -59,6 +59,8 @@ public abstract class IO {
   public static final String IGNORESUFFIX = ".ignore";
   /** File prefix. */
   public static final String FILEPREF = "file:/";
+  /** Jar prefix. */
+  public static final String JARPREF = "jar:";
 
   /** XQuery suffixes. */
   public static final String[] XQSUFFIXES =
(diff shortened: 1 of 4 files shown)


ChristianGruen:
Gunther, thanks for the pull request, very appreciated. I'll look into it soon.

ChristianGruen:
…diligent and stable, thanks for the comments.

GuntherRademacher:
Thank you, Christian!

PR #2155, opened 2023-01-04
Fix XXE vulnerability in XmlParser.java

sudosammy:
Thanks for your effort maintaining this library :)

In version 9.1 [this method](https://github.com/BaseXdb/basex/blob/9.1/basex-core/src/main/java/org/basex/io/parse/xml/XmlParser.java#L58) was added which cannot be configured safely to prevent XXE.

Regardless of the configured `dtd` bool, without control over the `external-general-entities` feature, external DTDs are evaluated leading to exploitation.

By my understanding of the library a user passing `false` to the `dtd` parameter would not want/expect external DTDs to be evaluated, therefore allowing the feature flag to be controlled by this method resolves the issue.


ChristianGruen:
Completely true; thanks for the fix.

PR #2156, opened 2023-01-09
Avoid long value overflow in position range

GuntherRademacher:
An XQuery positional filter for the last one or two elements of a sequence,
```xquery
    <a/>[position() >= last() - 1]
```
failed with this result, when there is just a single item in the context:
```
    [FOAR0002] Value out of range: 9223372036854775807.
```
This is caused by [Pos.get](https://github.com/BaseXdb/basex/blob/73986c1c4c156eea9c236e432d011fd00162bd8d/basex-core/src/main/java/org/basex/query/expr/Pos.java#L71-L77) transforming the filter predicate to a range ending in Long.MAX_VALUE,
```java
        case GE:
          minMax = new Expr[] { ps, Int.MAX };
          break;
        case GT:
          minMax = new Expr[] { new Arith(ii, type.instanceOf(AtomType.INTEGER) ? ps :
            cc.function(Function.FLOOR, ii, ps), Int.ONE, Calc.PLUS).optimize(cc), Int.MAX };
          break;
```
and [Range.value](https://github.com/BaseXdb/basex/blob/73986c1c4c156eea9c236e432d011fd00162bd8d/basex-core/src/main/java/org/basex/query/expr/Range.java#L72-L75) producing a long overflow while calculating the range size,
```java
    final long size = mx - mn + 1;
    if(size > 0) return RangeSeq.get(mn, size, true);
    // overflow of long value
    throw RANGE_X.get(info, mx);
```
where `mn == 0` and `mx == Long.MAX_VALUE`.

Actually the lower bound of the position range never needs to be less than 1, so the idea of this fix is to rewrite 
```xquery
    <a/>[position() >= last() - 1]
```
to
```xquery
    <a/>[position() >= max((1, last() - 1)]
```

### DIFF
--- a/basex-core/src/main/java/org/basex/query/expr/Pos.java
+++ b/basex-core/src/main/java/org/basex/query/expr/Pos.java
@@ -69,11 +69,11 @@ final class Pos extends Single {
           minMax = new Expr[] { ps, ps };
           break;
         case GE:
-          minMax = new Expr[] { ps, Int.MAX };
+          minMax = new Expr[] { cc.function(Function.MAX, ii, new List(ii, Int.ONE, ps).optimize(cc)), Int.MAX };
           break;
         case GT:
-          minMax = new Expr[] { new Arith(ii, type.instanceOf(AtomType.INTEGER) ? ps :
-            cc.function(Function.FLOOR, ii, ps), Int.ONE, Calc.PLUS).optimize(cc), Int.MAX };
+          minMax = new Expr[] { cc.function(Function.MAX, ii, new List(ii, Int.ONE, new Arith(ii, type.instanceOf(AtomType.INTEGER) ? ps :
+            cc.function(Function.FLOOR, ii, ps), Int.ONE, Calc.PLUS))).optimize(cc), Int.MAX };
           break;
         case LE:
           minMax = new Expr[] { Int.ONE, ps };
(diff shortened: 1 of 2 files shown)


ChristianGruen:
An excellent analysis! And a reasonable proposal to resolve the problem.

I noticed there are various other cases with minimum/maximum long values that cause questionable results or are rejected, e.g.:

```xquery
(: [FOAR0002] Value out of range: 9223372036854775807. :)
<a/>[position() = 0 to 9223372036854775807]
```

…so I decided to change the general behavior of [Range.value](https://github.com/BaseXdb/basex/blob/ce7a31ac4274c4b45348e8128d659fc8605ba329/basex-core/src/main/java/org/basex/query/expr/Range.java#L73): If an over-/underflow occurs, I’m now passing on the maximum long value instead of raising an error.

Admittedly, the solution introduces new errors. For example, the following query had been rejected, whereas it now returns a wrong result (`9223372036854775807` instead of `9223372036854775808`):

```xquery
count(0 to 9223372036854775807)
```

There would be several ways to fix this (use BigInteger instances for large integers; use flags/constants to indicate infinite values; etc.), but I don’t think it’s worth fixing that. The specification of the language is pretty vague when it comes to handling edge cases, and there appears to be no XQuery implementation that yields 100% correct results for such cases.

GuntherRademacher:
Thanks a lot!

PR #2158, opened 2023-01-10
Avoid OOME in SimpleTest 

GuntherRademacher:
Sorry for coming back to this, but there are two test cases in [SimpleTest](https://github.com/BaseXdb/basex/blob/main/basex-core/src/test/java/org/basex/query/simple/SimpleTest.java#L230-L231) that are now causing out-of-memory after the recent changes with respect to ranges:
```java
      { "Limits 11", "0 to 9223372036854775807" },
      { "Limits 11", "-9223372036854775807 to 9223372036854775807" },
```
I guess these just have to go now.


ChristianGruen:
You are so right. Thanks.

GuntherRademacher:
Thank you!

PR #2171, opened 2023-02-13
Suppress indentation adjacent to (X)HTML inline elements

GuntherRademacher:
More unwanted whitespace was found using (X)HTML serialization with `indent=yes`. An example of this is

```xml
<html><style type="text/css"><!--
body {font-size: 48px}
a {background: #CCC}
--></style><body><p><a name="x"><b>X</b></a> <a name="y"><b>X</b></a></p></body></html>
```
As is, browsers show it as

image

But when serialized with `method=html` and `indent=yes`, it becomes

```html
<html>
  <style type="text/css">
    <!--
body {font-size: 48px}
a {background: #CCC}
-->
  </style>
  <body>
    <p>
      <a name="x">
        <b>X</b>
      </a>
      <a name="y">
        <b>X</b>
      </a>
    </p>
  </body>
</html>
```

and browsers show it as

image

The difference in this case is caused by the indentation whitespace that was added between the closing tags for `b` and `a`. According to the serialization spec, that whitespace should not be there, because `a` and `b` are "inline" elements - see

* [6.1.4 XHTML Output Method: the indent and suppress-indentation Parameters](https://www.w3.org/TR/xslt-xquery-serialization-31/#XHTML_INDENT) 
* [7.4.3 HTML Output Method: the indent and suppress-indentation Parameters](https://www.w3.org/TR/xslt-xquery-serialization-31/#HTML_INDENT)

This fix keeps track of the most recent tag, in case it is a closing tag. When adding indentation, that and a following opening tag are checked for being inline elements. In case either of them is, the whitespace is suppressed.

As the spec suggests, the list of inline elements is taken from HTML and XHTML DTDs, and from HTML 5. Different lists could be used for HTML and XHTML, and in contrast to the implementation here, `ins` and `del` could additionally be checked for child elements before suppressing the whitespace. However for the sake of simplicity, I thought it is acceptable to use a single list and omit those extra checks. This would in the worst case add slightly less indentation whitespace than could be.

### DIFF
--- a/basex-core/src/main/java/org/basex/io/serial/HTMLSerializer.java
+++ b/basex-core/src/main/java/org/basex/io/serial/HTMLSerializer.java
@@ -23,6 +23,8 @@ final class HTMLSerializer extends MarkupSerializer {
   static final TokenSet EMPTIES5;
   /** (X)HTML: formatted elements. */
   static final TokenSet FORMATTEDS;
+  /** (X)HTML: inline elements. */
+  static final TokenSet INLINES;
   /** (X)HTML: URI attributes. */
   static final TokenSet URIS;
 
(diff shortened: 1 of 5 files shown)


GuntherRademacher:
Thanks! I will have a look at Checkstyle.

PR #2176, opened 2023-02-21
Do not add indentation whitespace following empty (X)HTML inline elements

GuntherRademacher:
After the recent changes of (X)HTML serialization for #2171, concerning inline elements, excess indentation whitespace could still be added following an empty tag of an inline element. This was caused by the empty tag not being remembered as the last closed tag.

This fix corrects the setting of `Serializer.closed` to that respect, and it extends HTML and XTHML serialization tests accordingly.


ChristianGruen:
…looks good.

PR #2181, opened 2023-02-27
Fix compile errors in mvn build

GuntherRademacher:
The mvn build failed, after recent changes of `Sandbox.java`: 

```log
[INFO] Compiling 1688 source files to C:\etc\github\GuntherRademacher\basex\basex-core\target\classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /C:/etc/github/GuntherRademacher/basex/basex-core/src/main/java/org/basex/core/Sandbox.java:[4,36] package org.junit.jupiter.api does not exist
```
A reference to junit had been added to Sandbox.java, but junit is not in scope when `src/main/java` is compiled.

The changes in this PR make the build work again by moving `Sandbox.java` to `src/test/java` and modifies `pom.xml` such that an additional test jar is created, which is then added as a dependency to `basex-api` and `basex-tests`.

This solution is tagged "[the easy way](https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html#the-easy-way)" in Maven documentation. The new test jar however has the same manifest as the primary jar: `Main-Class: org.basex.BaseXGUI`. This does not make sense, though it probably does not bother either. According to Maven documentation, "[the preferred way](https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html#the-easy-way)" would be to put base classes for tests into a separate project.


ChristianGruen:
…so this explains why we didn’t use JUnit methods in the Sandbox class (I didn’t remember). Good to know. It’s just consistent to move this class to the test directory, as there’s no need for it in the productive code. The easy way looks fine to me. Thanks!

PR #2187, opened 2023-03-13
Move Sandbox.java back to src/main/java

GuntherRademacher:
After introducing references to JUnit to `Sandbox.java` in 868ea59, the mvn build failed because JUnit was not in scope when compiling classes in `src/main/java`. This was fixed in d8c36be by moving `Sandbox.java` to `src/test/java`, creating a test jar in `basex-core`, and referring to it in `basex-api` and `basex-tests`. The test jar however contains all test classes from `basex-core`, so tests meant for `basex-core` are now also found when running tests for `basex-api` and `basex-tests`.

This fix reverts d8c36be, i.e. it moves `Sandbox.java` back to `src/main/java` and removes the test jar. JUnit references from `Sandbox.java` are now resolved by setting the scope of JUnit in `basex-core` to `provided`. This makes JUnit visible at compile time, without adding its artifacts to the build result.


ChristianGruen:
Works fine, and definitely a better solution than before, when `Util.notExpected` was used. Thanks.

PR #2192, opened 2023-03-27
Add support for fn:invisible-xml

GuntherRademacher:
# Add support for fn:invisible-xml

This change set adds `fn:invisible-xml` per Mike Kay's proposal (see [Support Invisible XML](https://github.com/qt4cg/qtspecs/issues/238) by @michaelhkay). The implementation is based on Norm Walsh's [CoffeeFilter](https://github.com/nineml/coffeefilter) and [CoffeeGrinder](https://github.com/nineml/coffeegrinder) projects (thank you, @ndw, for providing them).

Coffee*er jars are not planned to be delivered with BaseX, rather the user will have to take care of putting them in the runtime classpath. The implementation checks for the availability of the immediately referenced classes, and issues an error message in case they are unavailable. The code however will be compiled against those jars. In `pom.xml`, they are tagged as optional in scope `runtime`, and in `basex-core/pom.xml` the scope is `provided`, in order to have them available at compile time. With this setup, the jars will not be copied to `lib` folders and they will not appear in release artifacts.

Ixml parser generation is done in [FnInvisibleXml.java:67](https://github.com/GuntherRademacher/basex/blob/cc8d38b38470ccb06bb07b601a0036b973ff4849/basex-core/src/main/java/org/basex/query/func/fn/FnInvisibleXml.java#L67):

```java
      final InvisibleXmlParser parser = new InvisibleXml().getParserFromIxml(grammar);
```
Here, a new `InvisibleXml` object is created upon each invocation, though I think that it might be sufficient to have a single static instance of `InvisibleXml`. However after processing the first invalid grammar, all further parses will write some logging. I believe that this is not the intended behavior, so have asked @ndw to change it in nineml/coffeefilter#80.

The documentation should contain something like the following (though I am not sure where to put it, as I think XQuery 4 functions aren't documented yet, either):

## fn:invisible-xml

**Signature:**

```xquery
fn:invisible-xml(
  $grammar as xs:string
) as (function(xs:string) as document-node())
```
**Summary:**

The function takes as input a string defining an [invisible XML](https://invisiblexml.org/) grammar in ixml format, and returns as output a function that can be used to parse strings conforming to that grammar, converting them into XDM document nodes.

**Errors**:

`Failed to parse ixml grammar: could not match % at line %, column %.`

`Failed to generate ixml parser: %`

`Failed to parse ixml input: could not match % at line %, column %.`

`Failed to process ixml parser result: %`

**Prerequisites:**

The implementation of `fn:invisible-xml` relies on the [`org.nineml:coffeefilter`](https://github.com/nineml/coffeefilter) and [`org.nineml:coffeegrinder`](https://github.com/nineml/coffeefilter) projects. Their jars are not delivered with BaseX, but must be put into the runtime classpath explicitly. If one of their referenced classes is unavailable, an error message will be generated:

`Function invisible-xml requires missing class: org.nineml.%`


ChristianGruen:
> The documentation should contain something like the following (though I am not sure where to put it, as I think XQuery 4 functions aren't documented yet, either):

Thanks, that’s helpful. Before BaseX 11 is released, I’ll add an article about the current state of XQuery 4 (and BaseX-specific details) in our documentation, similar to the one for [XQuery 3.1](https://docs.basex.org/wiki/XQuery_3.1).

The official documentation of the function will also be added to the [XQFO 4.0 Specification](https://qt4cg.org/specifications/xpath-functions-40/Overview.html).

ChristianGruen:
@ndw That’s kind. It’s pretty convenient already to get the two jar files embedded. The main reason why we plan to make the library optional for the moment is that we want to keep our core library as small as possible.

PR #2198, opened 2023-04-08
Move getTree invocation back into try block

GuntherRademacher:
Sorry, my fault. The invocation of `getTree` had left the `try` block, but it must stay in there to have `IxmlException` caught and handled with `IXML_RESULT_X`. Obviously I had not run the tests...


ChristianGruen:
…thanks again.

PR #2255, opened 2023-10-28
Ignore encoding from XML declaration when parsing a String value

GuntherRademacher:
Some time ago I was using a Java tool (can't remember what it was) that generated XML in a Java string starting with an XML declaration of

```xml
    <?xml version="1.0" encoding="UTF-16"?>...
```

When passing that via `IO.get` to the `DBNode` constructor,

```java
    new DBNode(IO.get(xml))
```

it failed with a parsing error, because the string is internally encoded in UTF-8, the resulting byte stream is then passed to the XML parser, which decodes it per the encoding from the XML declaration. At the time I had fixed this in the application by omitting the XML declaration, but later I realized that the same is reproducible by a query like

```xquery
    parse-xml('<?xml version="1.0" encoding="utf-16"?><xml/>')
```

This PR is a proposal for fixing this, by making making `IOContent` supply a `Reader` that will decode the byte array as UTF-8, and have the XML parser ignore the encoding presented in the XML declaration.

### DIFF
--- a/basex-core/src/main/java/org/basex/build/xml/SAXWrapper.java
+++ b/basex-core/src/main/java/org/basex/build/xml/SAXWrapper.java
@@ -88,10 +88,33 @@ public final class SAXWrapper extends SingleParser {
    */
   @SuppressWarnings("resource")
   private InputSource inputSource() throws IOException {
+    length = source.length();
+
+    if (source.hasReader()) {
+      length = Math.max(0, length);
+      final Reader r = source.reader();
+      final InputSource rs = new InputSource(new Reader() {
+
+        @Override
+        public int read(final char[] cbuf, final int off, final int len) throws IOException {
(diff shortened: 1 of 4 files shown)


GuntherRademacher:
> This PR is a proposal for fixing this, by making making `IOContent` supply a `Reader` that will decode the byte array as UTF-8, and have the XML parser ignore the encoding presented in the XML declaration.

My first attempt to fix this was neglecting the fact that the encoding of `IOContent` can be different from UTF-8, as in `FetchModuleTest.binaryDoc`. Also for some reason I missed to use `InputSource.setEncoding`.

I have now replaced it by
  - keeping the encoding in `IOContent`, in case it was constructed from a `String`,
  - using `InputSource.setEncoding`.

ChristianGruen:
…merged (with just some minor changes).

PR #2256, opened 2023-11-03
Adapt fn:invisible-xml to draft specification

GuntherRademacher:
The [draft specification](https://qt4cg.org/pr/791/xpath-functions-40/Overview.html#func-invisible-xml) for `fn:invisible-xml` needs these changes in the implementation:

- an empty sequence for `$grammar` must use the Invisible XML specification grammar,
- the optional `$options` argument must be supported,
- errors codes must be `FOIX0001` and `FOIX0002`.


ChristianGruen:
Just fine, thanks

PR #2258, opened 2023-11-13
fn:parse-html and Validator.nu dependency

GuntherRademacher:
These changes

- replace the dependency on TagSoup by a dependency on Validator.nu's HTML parser,
- replace TagSoup options by Validator.nu options,
- use Validator.nu for implementing `html:parse`,
- add function `fn:parse-html`, based on `html:parse` (but, in contrast to that, without falling back to the XML parser in case Validator.nu is unavailable).

Some slight restructuring of `StandardFunc` and `FuncOptions` was necessary to be able to produce error `FODC0012`.


GuntherRademacher:
The latest changes to this PR re-integrate TagSoup, which remains to be the default HTML parser for everything except `fn:parse-html`, that using Validator.nu as the default. 

`HtmlOptions` integrates the options from both TagSoup and Validator.nu. Both for `fn:parse-html`, `html:parse`, and the other users of `HtmlParser` that supply `HtmlOptions`, the `method` option can be used to select the parser to be used. This will be TagSoup for optin values `xml` and `html`, and Validator.nu for option value `nu`.

Out of the 1379 QT4-tests for `fn:parse-html`, 1367 are now passing, and  12 still fail. These still need to be inspected in more detail.

ChristianGruen:
Thanks, everything fine. Next, we could tackle the options of `fn:html-parse` and decide which ones we want to support, or which ones we want to discuss in the qtspecs repository. For example, I am not sure whether `encoding` should be retained, but it could be useful for the still-tobe-added `fn:html-doc` function.

PR #2265, opened 2023-12-18
fn:format-number: add parameter for explicit format; use DecimalFormatSymbols to predefine formats for languages

GuntherRademacher:
These changes implement `fn:format-number` as specified in qt4cg/qtspecs#340:

- a fourth parameter `format` is added for supplying an explicit format,
- language identifiers of available locales are used to predefine a decimal format for a language. This is accomplished by getting the symbols from either `java.text.DecimalFormatSymbols`, or `com.ibm.icu.text.DecimalFormatSymbols`.

Also the remaining 7 test failures from [format-number.xml](https://github.com/qt4cg/qt4tests/blob/master/fn/format-number.xml) are fixed. It took two individual changes to do this:

- grouping separators in the fractional part were ignored
- the analysis for regular grouping was incomplete, in that i allowed digits making up more than one group at the beginning of the picture.


ChristianGruen:
Just merged (via a local branch).

The solution was fine! I’ve further unified the decimal-format handling by replacing the constant strings with the new `DecimalFormatOptions` class.

PR #2266, opened 2023-12-22
Support BCP 47 language tags in fn:format-integer, fn:format-date, fn:format-number

GuntherRademacher:
This is a follow-up change to the recent changes to `fn:format-[integer|date|number]`, which unifies language support such that all changed functions now support BCP 47 language tags.


ChristianGruen:
Thanks!

PR #2269, opened 2024-01-22
Respect ICU collator properties while comparing collation elements

GuntherRademacher:
The comparison of collation elements has been improved, now respecting the collator's properties.

This fixes most of the QT4 test cases that failed after adding ICU4J to the classpath, in particular those with name prefixes

- compare
- fn-contains
- fn-ends-with
- fn-starts-with
- fn-substring-after
- fn-substring-before
- defaultcolldecl
- fo-test-fn-contains
- fo-test-fn-starts-with
- fo-test-fn-ends-with
- fo-test-fn-substring-before
- fo-test-fn-substring-after

Some with the following prefixed are still failing, I am still working on those:

- format-time
- UCA

### DIFF
--- a/basex-core/src/main/java/org/basex/query/util/collation/UCACollation.java
+++ b/basex-core/src/main/java/org/basex/query/util/collation/UCACollation.java
@@ -1,11 +1,14 @@
 package org.basex.query.util.collation;
 
+import static com.ibm.icu.text.CollationElementIterator.*;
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Great

ChristianGruen:
One more test case you could look at: `compare-041`.

PR #2283, opened 2024-02-26
Add function coercion for maps and arrays

GuntherRademacher:
These changes add function coercion for maps and arrays by creating a new function with the required signature, that wraps a map or array function call along with a type check for the required target type:

```java
    final FuncCall funcCall = new DynFuncCall(ii, sc, this, param);
    final TypeCheck check = new TypeCheck(ii, sc, funcCall, ft.declType, true);
```

This fixes these QT4 test cases:

- `filter-007`
- `filter-008`
- `for-each-013`
- `for-each-014`
- `map-get-100`
- `xqhof20`
- `xqhof22`


ChristianGruen:
Locally merged.

PR #2287, opened 2024-03-04
Update Markup Blitz dependency to 1.3

GuntherRademacher:
1.3 brings a significant performance improvement and thread-safe parsing functions.


ChristianGruen:
🔝thanks

Issue #2300, opened 2024-05-29
enum: inspect:type(), prof:type()


ChristianGruen:
The type output of `inspect:type()` and `prof:type()` should be improved for enum types.

To achieve this, the cleanest approach could be to introduce an `ItemType` that encapsulates the enumeration values.

PR #2307, opened 2024-06-24
Replace AtomType.ENUM and SeqType.values by EnumType

GuntherRademacher:
Besides type inspection (#2300), there is at least one more problem with the current representation of `enum` types: `MapType.keyType`, as a `Type`, can currently only hold `AtomType.ENUM` and misses the values. Thus the following currently returns `true`, in lack of any values to be checked, but should return `false`:

```xquery
{'b' cast as enum('b'): 1} instance of map(enum('a'), item())
```

Also, a multi-value `enum` is defined as a union of single-value `enum`s, but we currently use different representations for either: multi-value `enum`s have multiple values in one `SeqType` instance, but a `ChoiceItemType` may spread them over several instances. This can cause instance tests to fail, e.g. with a wrong result of `false` in this case:

```xquery
fn($a as (enum('a')|enum('b'))) as item()* {$a} instance of fn(enum('a', 'b')) as item()*
```

These changes

- remove `AtomType.ENUM` and `SeqType.values`.
- convert `EnumValues` into an implementation of `Type` called `EnumType`.
- combine consecutive `EnumType`s before creating a `ChoiceItemType`. This solves the above mentioned problem with the function instance test.
- add parentheses around the result of `UnionTest.toString`. This aligns the result with `ChoiceItemType` notation and avoids misunderstanding when an occurrence indicator is appended later on.
- add a few tests.


ChristianGruen:
Very nice; the new organisation of enums is much cleaner. Thanks.

PR #2317, opened 2024-07-15
Add support for RecordTest

GuntherRademacher:
These changes provide the implementation of [RecordTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-record-test). This accomplished by adding a new class `RecordType extends MapType`.

This fixes 20 QT4 test cases with prefixes

- `array-members`
- `DynamicFunctionCall-R`
- `RecordTest`
- `Keywords-map-pair`
- `Keywords-array-members`

There are more test cases involving `RecordTest`s that are still failing. Those require coercion for records, which is not yet implemented.


ChristianGruen:
Excellent work. I hope to be able to review it this week.

PR #2351, opened 2024-12-04
Fix instance tests for wildcard nametests, union nodetests, and choice item types

GuntherRademacher:
The QT4 `subtyping` tests have reported a number of failures that called for these fixes:

  - type errors were reported while pre-evaluating `Closure`s at compile time for a function body that is never executed by a query. Reporting them made all QT4 test cases with prefix `subtyping` fail with type errors, because they use a construct similar to this:
    ```xquery
    declare variable $var as item()* external := '';
    function() as xs:integer {$var} instance of function() as xs:decimal
    ```
 - `element(p1:a)` was incorrectly not considered an instance of `element(p1:*)`
 - `element(p1:*|p2:*)` was incorrectly considered an instance of  `element(p1:*)`
 - `element(a|b)` was incorrectly not considered an instance of  `element(*:a|*:b)`
 - `element(a|b|c)` was incorrectly considered an instance of  `(element(a)|element(b))`

These have been added as tests to `TypeTest`.

Note that for running the `subtyping` tests, `fn:load-xquery-module` is needed. These changes here however are independent of it.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryParser.java
+++ b/basex-core/src/main/java/org/basex/query/QueryParser.java
@@ -3282,7 +3282,7 @@ public class QueryParser extends InputParser {
   private SeqType castTarget() throws QueryException {
     Type type;
     if(wsConsume("(")) {
-      type = choiceItemType();
+      type = choiceItemType().type;
     } else {
       final QNm name = eQName(sc.elemNS, TYPEINVALID);
       if(!name.hasURI() && eq(name.local(), token(ENUM))) {
@@ -3341,7 +3341,7 @@ public class QueryParser extends InputParser {
    */
   private SeqType itemType() throws QueryException {
     // choice item type
-    if(wsConsume("(")) return SeqType.get(choiceItemType(), Occ.EXACTLY_ONE);
+    if(wsConsume("(")) return choiceItemType();
 
     // parse annotations and type name
     final AnnList anns = annotations(false).check(false, false);
@@ -3565,7 +3565,7 @@ public class QueryParser extends InputParser {
    * @return item type
    * @throws QueryException query exception
    */
-  private Type choiceItemType() throws QueryException {
+  private SeqType choiceItemType() throws QueryException {
     final ArrayList<SeqType> types = new ArrayList<>() {
       @Override
       public boolean add(final SeqType st) {
@@ -3589,7 +3589,7 @@ public class QueryParser extends InputParser {
       }
     } while(wsConsume("|"));
     check(')');
-    return types.size() == 1 ? types.get(0).type : new ChoiceItemType(types);
+    return types.size() == 1 ? types.get(0) : new ChoiceItemType(types).seqType();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


ChristianGruen:
I wonder what the following instance check is supposed to return:

```xquery
function() as xs:integer { '' }
  instance of function() as xs:decimal
```

If it is legal to raise a type error (because the function body can never be converted to an integer), I believe it should also be legal to raise an error if an invalid value is specified in a variable declaration. What is your opinion on this?

If yes, it is probably the test code that should be revised.

GuntherRademacher:
> ```xq
> function() as xs:integer { '' }
>   instance of function() as xs:decimal
> ```
> If it is legal to raise a type error (because the function body can never be converted to an integer), I believe it should also be legal to raise an error if an invalid value is specified in a variable declaration. What is your opinion on this?

The spec says this: 

> [2.4.1 Kinds of Errors](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-kinds-of-errors)
> 
> As described in [2.3.3 Expression Processing](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-expression-processing), XQuery 4.0 defines a [static analysis phase](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-static-analysis), which does not depend on input data, and a [dynamic evaluation phase](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-dynamic-evaluation), which does depend on input data. Errors may be raised during each phase.
> 
> [Definition: An error that can be detected during the static analysis phase, and is not a type error, is a static error.] A syntax error is an example of a [static error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-static-error).
> 
> [Definition: A dynamic error is an error that must be detected during the dynamic evaluation phase and may be detected during the static analysis phase.] Numeric overflow is an example of a [dynamic error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-dynamic-error).
> 
> [Definition: A type error may be raised during the static analysis phase or the dynamic evaluation phase. During the static analysis phase, a [type error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-type-error) occurs when the [static type](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-static-type) of an expression does not match the expected type of the context in which the expression occurs. During the dynamic evaluation phase, a [type error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-type-error) occurs when the [dynamic type](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-dynamic-type) of a value does not match the expected type of the context in which the value occurs.]
> 

My interpretation was that a type error, which depends on the external variable value, must not be raised int the static analysis phase, because that is done independent of input data. In the dynamic evaluation phase, however, the dynamic type of the function body does not contribute to the query result, so it is not being evaluated and the error will not be raised. 

So the question is whether it is legitimate to intermix the static and dynamic context by considering known fixed values of external variables already in the static analysis phase. I could not find anything in the spec about this.

In any case, the behavior of dynamic functions, per the change in this PR, corresponds to that of static functions. Those are never reporting type errors that refer to the function body, unless that is being executed. Both of the queries below are succeeding: 

```xquery
declare function local:f() as xs:integer { '' };
local:f#0 instance of function() as xs:decimal
``` 

```xquery
declare variable $x external := '';
declare function local:f() as xs:integer { $x };
local:f#0 instance of function() as xs:decimal
```

PR #2357, opened 2024-12-17
Option value atomization

GuntherRademacher:
In a [recent change](https://github.com/BaseXdb/basex/commit/e961a87040900e46c9ee5f6eb1a1f1f0cf405347#diff-184e184e99b0eee9d6d7f02bce510d6bf7abeea9cc0525619837005b0a1c161bL615-L640), the atomization of option values was removed to allow passing values of type `ValueOption` without being atomized. However, this inadvertently caused `StringsOption` values to no longer accept arrays, and the issue was not detected by existing tests.

I now assume that the atomization was meant to implement this part of the [option parameter conventions](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#option-parameter-conventions):

> In cases where an option is list-valued, by convention the function should accept either a sequence or an array: but this rule applies only if the specification of the option explicitly accepts either. Accepting a sequence is convenient if the value is generated programmatically using an XPath expression; while accepting an array allows the options to be held in an external file in JSON format, to be read using a call on the fn:json-doc function.

Though, I could not find any option making use of this:

> ...this rule applies only if the specification of the option explicitly accepts either...

But I think the JSON compatibility note makes sense, and it could well apply to all list-valued options. Does it? If yes, the above clause should possibly be dropped from the spec.

This change now reinstalls atomization to `StringsOption` handling, and adds a test for it.

In addition, for a `StringOption`, a check is added that explicitly rejects a sequence.

### DIFF
--- a/basex-core/src/main/java/org/basex/build/csv/CsvOptions.java
+++ b/basex-core/src/main/java/org/basex/build/csv/CsvOptions.java
@@ -74,9 +74,9 @@ public class CsvOptions extends Options {
   }
 
   @Override
-  public synchronized void assign(final Item name, final Value value, final InputInfo info)
-      throws QueryException {
-    super.assign(name, value, info);
+  public synchronized void assign(final Item name, final Value value, final QueryContext qc,
+      final InputInfo info) throws QueryException {
+    super.assign(name, value, qc, info);
     if(separator() == -1) throw OPTION_X.get(info, "Invalid separator: '%'", get(SEPARATOR));
   }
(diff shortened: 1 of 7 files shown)


ChristianGruen:
As shortly discussed, it may be better to stick with the status quo for now and look at the coercion of maps more broadly later on. 

In issue 1661 of qtspecs, I proposed to make the parsing of QName options more flexible. The consequences may also affect our option handling.

PR #2375, opened 2025-01-30
Update workflow action versions

GuntherRademacher:
Coincidentally `v3` of the actions used in the build workflow went out of service today, which caused a [recent execution](https://github.com/BaseXdb/basex/actions/runs/13057309382) to fail. So we need to update to `v4`.


ChristianGruen:
Just in time, thanks ;·)

PR #2376, opened 2025-02-04
CSV module: restore 'yes'/'no' values for header option, handle explicit column names

GuntherRademacher:
After a recent change, it was no longer possible to use `'yes'`/`'no'` or `'1'`/`'0'` instead of `true()`/`false()` as values of the `header` option of functions in the CSV module.

These changes restore this capability. In addition, they also support explicit column names in the `header` option, except for single strings with one of the above values.


ChristianGruen (approved):
Just fine.

PR #2377, opened 2025-02-04
Use Validator.nu as default parser when TagSoup is unavailable

GuntherRademacher:
This change will use Validator.nu as the default HTML parser, when TagSoup is unavailable, or fall back to XML parsing, if Validator.nu is unavailable, too.

When `HTMLPARSER` option `method` is used to override the default, then an error message is produced, if the specified parser is unavailable.


GuntherRademacher:
The documentation has been updated to explain this behavior: https://docs.basex.org/12/Parsers

PR #2381, opened 2025-02-12
Add options to `fn:parse-xml` and `fn:parse-xml-fragment`

GuntherRademacher:
These changes add the `$options` parameter to `fn:parse-xml` and `fn:parse-xml-fragment`.

Besides `base-uri` and `strip-space`, the XML parsing options from `MainOptions` are also supported. The default values for those options are taken from `qc.context.options`.

`dtd-validation` and `xsd-validation` are defined in `ParseXmlFragmentOptions`, but otherwise ignored.

### DIFF
--- a/basex-core/src/main/java/org/basex/build/Parser.java
+++ b/basex-core/src/main/java/org/basex/build/Parser.java
@@ -116,7 +116,17 @@ public class Parser extends Job {
    * @return xml parser
    */
   public static SAXWrapper xmlParser(final IO source) {
-    return new SAXWrapper(source, new MainOptions());
+    return xmlParser(source, new MainOptions());
+  }
+
+  /**
+   * Returns an XML parser instance, using the Java default parser.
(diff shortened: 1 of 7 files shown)


ChristianGruen:
All fine! My only changes:

* I have removed the assignment of global parsing options in accordance with other standard functions, like `fn:serialize` that uses the standard serialization method `xml` instead of the BaseX default method `basex`. This is to ensure that the functions always behave identically, no matter how the current global state looks like.
* I have renamed `strip-ns` to `stripns` and `int-parse` to `intparse`. I believe this will simplify things: In the documentation, we can mention that the existing [Parsing Options](https://docs.basex.org/12/Options#parsing) can be supplied for those options that have no no standard counterpart.

As you may have noticed, the [Fetch Functions](https://docs.basex.org/12/Fetch_Functions) provide similar functionality. Some of them could become obsolete in a future version of BaseX.

PR #2382, opened 2025-02-19
qt1528: Implement constraint "unreserved name"

GuntherRademacher:
This change implements [constraint "unreserved name"](https://qt4cg.org/specifications/xquery-40/xquery-40.html#extra-grammatical-constraints), by disallowing keywords as plain names in computed node constructors.

The specification says about this:

>  To ease transition, implementations may provide an option to allow such names to be accepted with a warning that the construct is deprecated.

If you think that we need such an option, please let me know.


ChristianGruen:
Thanks a lot for the evaluation of the conflicting keywords in https://github.com/qt4cg/qtspecs/issues/1528#issuecomment-2887716022.

I’m closing the PR, as it will be outdated and as it is more or less identical to my internal one.

PR #2394, opened 2025-03-04
Support XML representation of grammar in `fn:invisible-xml`

GuntherRademacher:
`fn:invisible-xml` now accepts an XML representation of the grammar as the first parameter.


ChristianGruen:
Great

PR #2401, opened 2025-03-19
Method annotation

GuntherRademacher:
This change implements the [`%method`](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-methods) annotation for function items.

The context of a `%method` function item is bound to the relevant map from within `Lookup.valueFor`. Some optimizations have to be omitted if there is a possibility that the context binding must take place.


ChristianGruen:
Merged, works fine (and I have revised the flawed `mayBeFunction` test).

PR #2402, opened 2025-03-20
Java 13 compatibility for `FileSystems.newFileSystem`

nverwer:
Since Java 13 there are two `FileSystems.newFileSystem` methods which accept a null second parameter. This makes a call with `null` ambiguous. See [https://errorprone.info/bugpattern/NewFileSystem].

The proposed change casts the null parameter to the intended type.


ChristianGruen:
Thanks, appreciated.

PR #2404, opened 2025-03-23
Fix `fn:document-uri` for documents obtained from `fn:collection`

GuntherRademacher:
As reported by [Yitzhak Khabinsky on basex-talk](https://www.mail-archive.com/basex-talk@mailman.uni-konstanz.de/msg16100.html), `fn:document-uri` in 11.8 returns incomplete results, when applied to documents obtained from `fn:collection`. The issue was introduced by commit 83acbb74f9419d50a4c03c5793647d449b5577c1, which attempted to fix the duplication of path elements in cases where the original URI ends with a slash. For example,

```xquery
base-uri(doc('https://www.w3.org/TR/xml/'))
```

Up to 11.7, this used to return `https://www.w3.org/TR/xml/xml`, with an extra `xml` at the end.

That change was made under the assumption that `MetaData.original` already contained the full path of the document, which however does not hold for `fn:collection` documents.

The change in this PR now adds a new flag `MetaData.single`, providing the information whether a `MetaData` instance is used for a single document. If the flag is set, `MetaData.original` contains the full path to be returnd by `DBNode.baseURI`, otherwise it must be combined  with the name stored in the document.

I have also explored other approaches to fix the issue, but without success. The name stored in the document originates from `IO.nm`, and I tried to set this to an empty token for paths ending in a slash. While this resolved the specific issue, it caused numerous failures in other tests.

### DIFF
--- a/basex-core/src/main/java/org/basex/data/MetaData.java
+++ b/basex-core/src/main/java/org/basex/data/MetaData.java
@@ -30,6 +30,8 @@ public final class MetaData {
 
   /** Path to initially imported resources. */
   public String original = "";
+  /** Whether this instance is for a single document. */
+  public boolean single = true;
   /** Size of initially imported resources. */
   public long inputsize;
   /** Database timestamp. */
--- a/basex-core/src/main/java/org/basex/query/QueryResources.java
+++ b/basex-core/src/main/java/org/basex/query/QueryResources.java
@@ -473,6 +473,7 @@ public final class QueryResources {
     final Data data;
     try {
       data = CreateDB.create(io.dbName(), parser, context, opts, mem);
+      data.meta.single = single;
     } catch(final IOException ex) {
       throw IOERR_X.get(info, ex);
     }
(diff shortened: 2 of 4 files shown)


ChristianGruen:
Maybe `MetaData.ndocs` could be considered instead of adding a new variable? But I’ll have a closer look tomorrow.

GuntherRademacher:
I have checked all elements of `MetaData`, but have not yet found one that is suitable. `ndocs` is `1` both for a single document retrieved with `fn:doc` and for a folder containing only one document retrieved with `fn:collection`.

ChristianGruen:
> I have checked all elements of `MetaData`, but have not yet found one that is suitable. `ndocs` is `1` both for a single document retrieved with `fn:doc` and for a folder containing only one document retrieved with `fn:collection`.

Oh yes; thanks.

I ended up taking a different approach to fixing this, although I have to admit it may not be a completely universal one:
* For documents created with `fn:parse-xml`, `IOContent` instances are created, which have internal `name` and `path` references.
* If the base URI points to a directory, `IOContent.name` is now set to an empty string.
* As a result, the original path to the base URI (`data.meta.original`) remains unchanged in the resulting document, but the specific URI of the document (which is retrieved via `data.text(0, true)` will be empty.
* This way, it cannot happen anymore that segments in the resulting URI will be duplicated.

To make this fully consistent, much more spent would probably need to be spent on the convenience `IO` classes (which serve all kinds of purposes).

Give me a note if you doubt that this fix covers all cases you observed.

PR #2411, opened 2025-04-04
Implement recent changes in regular expression handling

GuntherRademacher:
This change implements the following changes of regular expression handling

- word boundary assertions
- lookahead and lookbehind assertions
- representation of lookahead groups in `fn:analyze-string` results
- removal of matches-empty restriction for `fn:analyze-string`, `fn:tokenize`, `fn:replace`


ChristianGruen:
…excellent

PR #2422, opened 2025-04-23
New options for `fn:doc`, `fn:doc-available`, and `fn:parse-xml`

GuntherRademacher:
These changes implement the new options for `fn:doc`, `fn:doc-available`. When fetching non-database documents, the functions no longer make use of the main options, rather the new `DocOptions` class provides the available options and default values. But providing explicit options is only supported for non-database documents, while it is rejected, when the the path turns out to identify a database document. Explicit options are also rejected, when the functions are used to create database documents by using `FORCECREATE`.

Option `allow-external-entities` is related, but different to existing option `dtd`. A value of `false()` for `dtd` asks for ignoring external entities, while `false()` for `allow-external-entities` will cause failure for any reference of an external entity.

`MetaData` has been extended with a new member `docOpts`. It is populated for non-database documents only, and it helps creating the same node, when identical documents are loaded more than once with equal options.

This also adds the new options of `fn:parse-xml`.

In contrast to `fn:parse-xml`, `fn:doc` and `fn:doc-available` have not been equipped with option `catalog`. The reason for this was the structure of the code in `Docs`, where `queryInput` is evaluated early, before the options have been looked at. But in order to honor `catalog`, the options would have needed to be processed before evaluating `queryInput`. Rather `catalog` is now taken from `context.options`. Actually I was not sure whether this the preferable alternative, or whether `catalog` should be part of the function's options as it was done for `fn:parse-xml`. This should be done consistently in the same way in either case -  please let me know which way to go.

One problem that was detected while working on these changes: documents loaded by `fn:doc` and `fn:collection` will create different nodes, even when identical. This is because of the comparison `IO.get(original).eq(qi.io)` in `QueryResources.data`, it fails because `fn:collection` and `fn:doc` cause different content here. A QT4 test for this, `collection-009`, does not detect the problem, because in the QT4 test environment, the collection documents are pre-loaded before the actual query execution.

The changes pass all QT4 tests with prefixes `fn-doc-` and `parse-xml`.

### DIFF
--- a/basex-core/src/main/java/org/basex/build/xml/SAXWrapper.java
+++ b/basex-core/src/main/java/org/basex/build/xml/SAXWrapper.java
@@ -54,8 +54,10 @@ public final class SAXWrapper extends SingleParser {
       if(reader == null) {
         reader = XmlParser.reader(options);
       }
-      final EntityResolver er = Resolver.entities(options);
-      if(er != null) reader.setEntityResolver(er);
+      if(reader.getEntityResolver() == null) {
+        final EntityResolver er = Resolver.entities(options);
+        if(er != null) reader.setEntityResolver(er);
+      }
 
       saxh = new SAXHandler(builder, options.get(MainOptions.STRIPWS),
           options.get(MainOptions.STRIPNS));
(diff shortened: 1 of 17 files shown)


ChristianGruen:
Merged, after some local refactorings. Thank you for the comprehensive summary in the PR (I’ll give mine later on).

PR #2458, opened 2025-07-15
Add support for built-in named record types

GuntherRademacher:
These changes add support for [built-in named record types](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#id-built-in-named-record-types). These types are now defined in `SeqType`. Function signatures have been adapted to use them, and the order of generated map elements has been aligned with the type definition.

Constructor functions for built-in record types are handled in the same way as user-defined record type constructors, i.e. they are generated dynamically from the record definition. While this makes sure that they match the record definition, it is a little unfortunate, because they are created for each and every query. This currently must be done in order to make them available for `fn:function-lookup`. I would expect these functions to be rarely used, if ever, so it would be desirable to save the effort for creating them dynamically over and over again. I am thus thinking of replacing the dynamic generation with fixed Java implementations, or possibly finding a way to generate them only once.

This fixes a total of 39 QT4 tests. Two specific tests, `built-in-record-type-303` and `built-in-record-type-304` do not succeed yet, because of a problem with function item types: in [`Closure.optimize(CompileContext)`](https://github.com/BaseXdb/basex/blob/ad5301568536465450e9db061a2b3b5d18d2a947/basex-core/src/main/java/org/basex/query/func/Closure.java#L196-L198), the result type is inferred from the function body:

```java
    final SeqType st = expr.seqType();
    final SeqType dt = declType == null || st.instanceOf(declType) ? st : declType;
    exprType.assign(FuncType.get(anns, dt, params));
```

This makes the following `instance of` succeed, but it is expected to fail:

```xquery
fn() as xs:anyAtomicType? {'banana'} instance of function() as xs:string
```

It could be fixed replacing the above three lines with just

```java
    exprType.assign(FuncType.get(anns, declType, params));
```

But while this change is fine with respect to QT4 tests, it causes a number of BaseX tests to fail, and some hang because the code is optimized differently. So this needs further investigation as well.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryParser.java
+++ b/basex-core/src/main/java/org/basex/query/QueryParser.java
@@ -275,6 +275,11 @@ public class QueryParser extends InputParser {
    * @throws QueryException query exception
    */
   private void check(final MainModule main) throws QueryException {
+    // add record constructor functions for built-in record types
+    for(final RecordType rt : SeqType.BUILT_IN_NAMED_RECORD_TYPES.values()) {
+      if(qc.functions.get(rt.name(), rt.minFields()) == null) declareRecordConstructor(rt, info());
+    }
+
     // check function calls and variable references
     qc.functions.check(qc);
     qc.vars.check();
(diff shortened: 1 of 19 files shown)


ChristianGruen:
@GuntherRademacher As you have indicated in your comment, it is not ideal to always create constructors, in particular because they are never used. Let’s see whether we will keep them in the spec (qtspecs, issue 2101).

Do you think a simplified of this PR could be created that only adds the new record types, and omits the code for providing the constructors?

Unfortunately, my recent commits have created some conflicts.

ChristianGruen:
PS: I have re-added the record constructors (619747a21ef29bd5924ccd9d6f925905adaa5d76); they were much more lightweight than I had expected.

PR #2492, opened 2025-08-20
Drop position from fold callbacks

GuntherRademacher:
This changes removes the position parameters form fold functions, as per qt4cg/qtspecs#1867.


ChristianGruen:
I have decided to discard the PR. Contrary to the official specification, it will still be possible in BaseX to retrieve the current position of the iteration via the third parameter of the higher-order function argument.

I have updated our docs:
* https://docs.basex.org/main/Standard_Functions#fn:fold-left
* https://docs.basex.org/main/Array_Functions#array:fold-left.

PR #2499, opened 2025-09-17
Add Turkish language support for BaseX

selcukgural:
Hello,

I tried to translate it into Turkish as much as possible, and I also took care to keep some words in English because some words in Turkish do not really have a fixed, universally used equivalent.


ChristianGruen:
Thanks for the Turkish translation, much appreciated.
Feel free to add your name or alias in line 2.

PR #2503, opened 2025-09-24
Allow a sequence of function items in dynamic function calls

GuntherRademacher:
These changes add support for [PR 1975](https://github.com/qt4cg/qtspecs/pull/1975), i.e. allowing dynamic function calls to operate on a sequence of function items, fixing the corresponding QT4-Tests. Unfortunately there are  3 errors in BaseX-Tests, because tail-call optimization for dynamic function calls does not work any longer:
 - [RewritingsTest.gh2314](https://github.com/BaseXdb/basex/blob/d88855a4f2113315859204b96016b9fa46d23d5e/basex-core/src/test/java/org/basex/query/ast/RewritingsTest.java#L3364-L3367)
 - [TCOTest.dynFuncCall](https://github.com/BaseXdb/basex/blob/d88855a4f2113315859204b96016b9fa46d23d5e/basex-core/src/test/java/org/basex/query/ast/TCOTest.java#L108-L123)
 - [FuncItemTest.funcItemInlining](https://github.com/BaseXdb/basex/blob/d88855a4f2113315859204b96016b9fa46d23d5e/basex-core/src/test/java/org/basex/query/ast/FuncItemTest.java#L159-L179)

The approach taken here is to generate in iteration from within the parser. Effectively,
```xquery
(upper-case#1, lower-case#1)('aBcD')
```
 is translated to
```xquery
for $fn in (upper-case#1, lower-case#1) return $fn('aBcD')
```
but there also is special treatment without an interation for the case where the static type indicates exactly one function item.

An alternative approach might be enhancing the `DynFuncCall` operator to support a sequence of function items, but that has not been investigated yet.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryParser.java
+++ b/basex-core/src/main/java/org/basex/query/QueryParser.java
@@ -2506,7 +2506,7 @@ public class QueryParser extends InputParser {
         } else if(consume("=?>")) {
           expr = methodCall(expr);
         } else if(current('(')) {
-          expr = Functions.dynamic(expr, argumentList(false, null));
+          expr = dynamicFunctionCall(expr);
         } else if(current('?')) {
           expr = lookup(expr);
           if(expr == null) break;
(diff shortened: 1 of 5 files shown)


ChristianGruen:
I close with PR as I believe I have found an alternative solution (https://github.com/BaseXdb/basex/commit/6acd7ddd2c15f313f6a5b248603ff31c58c5105d) that does not endanger our TCO rewritings, and that will work for other syntax constructs as well.

PR #2513, opened 2025-10-15
Add support for functions in no namespace

GuntherRademacher:
These changes adapt to the new rules allowing functions in no namespace.

These rules require that unprefixed function names cannot be resolved before all user-defined functions have been declared. This is accomplished by new class `FuncRef`, acting as a placeholder for a static function call or named function item, delaying the construction of the implementation until its `resolve` method is called.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/func/FuncRef.java
+++ b/basex-core/src/main/java/org/basex/query/func/FuncRef.java
@@ -1,0 +1,140 @@
+package org.basex.query.func;
+
+import java.util.function.*;
+
+import org.basex.query.*;
+import org.basex.query.expr.*;
+import org.basex.query.iter.*;
+import org.basex.query.util.*;
+import org.basex.query.util.parse.*;
+import org.basex.query.value.*;
+import org.basex.query.value.item.*;
+import org.basex.query.value.type.*;
+import org.basex.query.var.*;
+import org.basex.util.*;
+import org.basex.util.hash.*;
+
+/**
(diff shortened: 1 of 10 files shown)


basex-core/src/main/java/org/basex/query/func/FuncRef.java
```diff
@@ -0,0 +1,140 @@
...
+    return expr().has(flags);
+  }
+
+  @Override
+  public boolean vacuous() {
+    return expr().vacuous();
+  }
+
+  @Override
+  public Iter iter(final QueryContext qc) throws QueryException {
```
ChristianGruen: We should try to get rid of all runtime references to `FuncRef`. For example, in the following expression that’s not the case yet:

```xquery
copy $c := parse-xml(<?_ <x/>?>)
modify ()
return $c
```

This one works: `parse-xml(<?_ <x/>?>) update {}`.

ChristianGruen:
A nice solution!

* I extended `FuncRef` from `Single`, and I removed various `Util.notExpected` calls (trusting that we know why possible NPEs are raised).
* I replaced `ThrowingFunction` with our similar `QueryFunction` interface.
* I fixed a bug in a related commit in the `Transform` class to prevent `FuncRef.iter` from being called at runtime (maybe there are more cases? We will see).

The `toString` solution is elegant. I wonder, however, if we (still) need it, or if we can replace it with `expr.toString(qs)`? Can you remember a case in which the function was invoked, and in which `expr` has not been initialized yet?

PR #2522, opened 2025-10-28
Add caching for `fn:load-xquery-module` 

GuntherRademacher:
These changes add caching for `fn:load-xquery-module` to prevent redundant retrieval and compilation of modules within the same query execution.

There are two new caches in `FnLoadXQueryModule`:

- Compiled module cache (`MOD_CACHE`)
  - `WeakHashMap<QueryContext, ThreadLocal<Map<CacheKey, XQMap>>>`
  - for each top-level `QueryContext`, holds a thread-local cache of compiled results
  - ensures that compiled functions and variables are reused within a thread but not shared across threads

- Source snapshot cache (`SRC_CACHE`)
  - `WeakHashMap<QueryContext, ConcurrentHashMap<String, String>>`
  - shared among threads of the same top-level query
  - prevents repeated I/O by reusing the same module source text when multiple threads access the same URI

The `CacheKey` for the compiled module cache is based on the module URI and the original options map (an `XQMap`). It uses deep structural equality and order-independent hashing of nested maps and arrays.

The  `COMPILED_MODULES` counter tracks the number of actual compilations (cache misses). This is used for testing only.

### DIFF
--- a/basex-api/src/main/webapp/dba/databases/database.xqm
+++ b/basex-api/src/main/webapp/dba/databases/database.xqm
@@ -46,7 +46,7 @@ function dba:database(
   let $db-exists := db:exists($name)
   return (
     <tr>{
-      <td>
+      <td width='33%'>
         <form method='post' autocomplete='off'>
           <input type='hidden' name='name' value='{ $name }' id='name'/>
           <h2>{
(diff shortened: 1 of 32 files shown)


GuntherRademacher:
> Maybe all cached instances could be organized via the [QueryThreads](https://github.com/BaseXdb/basex/blob/main/basex-core/src/main/java/org/basex/query/QueryThreads.java) class.

This has now been done in the latest commit,

PR #2603, opened 2026-02-18
Remove per-query generation of built-in record type constructors

GuntherRademacher:
This PR removes the per-query generation of built-in record type constructors in `QueryParser`. Instead, constructors for built-in record types are now registered once as standard functions via the `Function` enum.

The constructor signatures and parameter types are still derived from the `Records` enum (i.e., `RecordType` remains the single source of truth). A new `Function(RecordType)` constructor and corresponding helper methods in `RecordType` are used to generate:

- the function signature string (e.g., `name(a,b[,c]))`, and
- the constructor parameter types (in declaration order).

At runtime, the constructor implementation creates either:

- an `XQRecordMap` (for record types without optional fields), or
- a regular `XQMap` (if optional fields are present).

This change eliminates the repeated initialization overhead per query execution while preserving existing semantics.

There is now some similarity between `RecordType.constructor()` and `CRecord`. I did not reuse `CRecord`, as the `Function` infrastructure requires a `StandardFunc`. However, `CRecord` could potentially be refactored to build upon `RecordType.constructor()`. I’m submitting this early as I’ll be unavailable for the next few days, but I’m happy to follow up on the `CRecord` refactoring afterwards.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/func/Records.java
+++ b/basex-core/src/main/java/org/basex/query/func/Records.java
@@ -2,7 +2,6 @@ package org.basex.query.func;
 
 import static org.basex.query.QueryText.*;
 
-import org.basex.query.expr.*;
 import org.basex.query.util.hash.*;
 import org.basex.query.util.list.*;
 import org.basex.query.value.item.*;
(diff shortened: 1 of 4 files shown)


ChristianGruen:
Thanks, looks fine.

> However, `CRecord` could potentially be refactored to build upon `RecordType.constructor()`.

I agree this sounds promising.

PR #2627, opened 2026-03-25
Introduce two-pass processing for direct attribute parsing

GuntherRademacher:
This PR introduces a two-pass strategy for processing attribute lists in direct element constructors.

In the first pass, only `xmlns` attributes are processed, updating the namespace context. In the second pass, all remaining attributes are parsed. This ensures that namespace declarations are available to embedded expressions, even if those expressions precede the corresponding `xmlns` attributes.

An alternative approach based on deferred QName resolution (via `QNmResolver`) was considered. While this works for static function names (which are resolved after parsing), it does not extend well to:

- variable names, especially scoped local variables
- type names, which would require additional indirection similar to named record types

The two-pass strategy avoids these issues entirely and allows QName resolution to occur immediately. As a result, `QNmResolver` is no longer needed and has been removed.

The main challenge is skipping attribute values during the first or second pass without triggering parsing side effects. Direct attribute values can contain complex nested constructs, including enclosed expressions, literals, comments, direct constructors, pragmas, and templates. Reusing existing parsing code for skipping is not feasible due to its side effects. Instead, skipping is implemented using a parser generated from a simplified grammar derived from the XQuery 4.0 specification.

There are 3 commits in this PR:

- 332fe2b: introduce two-pass processing with REx-based attribute value scanning
- f4cc284: remove `QNmResolver` and switch to immediate QName resolution
- 8665f67: add regression tests

This change fixes the following QT4 test cases:

- `union-type-009`
- `union-type-010`
- `Constr-namespace-30`
- `K2-DirectConElemNamespace-8`
- `K2-DirectConElemNamespace-14`
- `K2-DirectConElemNamespace-19`
- `K2-DirectConElemNamespace-20`
- `K2-DirectConElemNamespace-21`
- `K2-DirectConElemNamespace-22`
- `K2-DirectConElemNamespace-29`
- `K2-DirectConElemNamespace-30`
- `K2-DirectConElemNamespace-31`
- `K2-DirectConElemNamespace-32`
- `K2-DirectConElemNamespace-33`
- `function-decl-40-035`
- `function-decl-40-036`
- `K2-NamespaceProlog-13`
- `try-catch-err-code-variable-13`


ChristianGruen:
I eventually decided to fix this with a self-contained alternative. A clear drawback is that it will require more time to fix it when the grammar changes, but for now I wanted to keep the number of lines small. The PR was very helpful, though: I adopted various parts of it, like the dissolution of the QNmResolver.

A promising long-term objective could be to use REx to parse the full grammar and to utilize the resulting structures in other parts of BaseX such as the GUI editor.

PR #2647, opened 2026-04-21
Add check for `xml:`-prefixed name to DB wildcard nametests

GuntherRademacher:
For wildcard name tests on database nodes backed by `DiskData`, `NameTest.optimize` can reduce the test to empty if the local name is not found in the data’s name table. However, names in the `xml` namespace are stored there with the `xml` prefix, so a lookup by local name alone does not find them. This change checks for the `xml:` -prefixed form before optimizing the result to empty.


ChristianGruen:
Thanks for the immediate fix of this longtime bug.

PR #2670, opened 2026-05-20
Add `fn:regex` and `fn:compiled-regex-record`

GuntherRademacher:
These changes add support for the new `fn:regex` function and the built-in `fn:compiled-regex-record` type.

`fn:regex` is implemented by constructing a map of `FuncItem` closures over the constant pattern and flags. Each closure body is a fresh `StandardFunc` instance (`FnMatches`, `FnTokenize`, etc.). Prior to this change, each such instance had its own per-instance regex cache. The per-instance cache has been replaced by a single per-query cache on `QueryContext`, shared by all `RegExFn` instances within a query. The pattern compiled eagerly in `fn:regex` is therefore found immediately by any subsequent member function call - fulfilling the spec's intent of compiling once and reusing.

`RegExpr`, `GroupInfo`, and `GroupScanner` were moved from inner classes of `RegExFn` to a new top-level class `org.basex.query.util.regex.RegExpr`. This was done in order to avoid `QueryContext` depending on a type defined inside a concrete function implementation. The only code changes are visibility adjustments required by the package move: the class, its constructor, the pattern field, and the two accessor methods are now public (they were package-private, but callers in `org.basex.query.func.fn` and `org.basex.query` require access), algorithmic logic is identical.

This fixes QT4 tests

- `Keywords-fn-regex-1`
- `fo-test-fn-regex-001`
- `fo-test-fn-regex-002`
- `fo-test-fn-regex-004`

Additionally, a test has been added to verify that constructor functions exist for all built-in record types, and constructor functions have been added where they were missing.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryContext.java
+++ b/basex-core/src/main/java/org/basex/query/QueryContext.java
@@ -26,6 +26,7 @@ import org.basex.io.serial.*;
 import org.basex.query.ann.*;
 import org.basex.query.expr.*;
 import org.basex.query.func.*;
+import org.basex.query.util.regex.*;
 import org.basex.query.func.java.*;
 import org.basex.query.iter.*;
 import org.basex.query.scope.*;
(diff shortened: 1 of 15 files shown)


GuntherRademacher:
A deliberate deviation from the spec: the `replace` member of `fn:compiled-regex-record` is typed as `xs:string` rather than `xs:string*`. The spec's `xs:string*` seems incorrect, `fn:replace` always returns exactly one string. So the stricter type is used here accordingly. A remark about this has been added to qtspecs#2590.

PR #2676, opened 2026-05-26
Add support for option `number-format` of `fn:parse-json`

GuntherRademacher:
These changes add support for option `number-format` to

- `fn:parse-json`
- `fn:json-doc`
- `json:parse`

It replaces option `number-parser` for these functions. `number-parser` continues to be supported for `fn:json-to-xml`. 

All of these options are defined in `JsonParserOptions`, though the allowances are different for different functions (this was even the case before these changes with option `null`). Upon invocation, there is an extra check for the option being supported with the particular format, but as it is now, it can only trigger an error upon non-default-values.

There are 14 QT4 test cases using `number-parser` for `fn:parse-json` and `fn:json-doc`, which are now failing because `number-parser` is rejected.


ChristianGruen:
…perfect.

PR #2688, opened 2026-06-03
Fix concurrent evaluation of static variables used from `xquery:fork-join`

GuntherRademacher:
Static variables are evaluated at most once per query, involving a circular dependency check, but this was triggered incorrectly when variable values were accessed and evaluated concurrently by multiple threads started via `xquery:fork-join` (see #2678).

The changes here

- replace the `dontEnter` runtime check with query-local static-variable evaluation coordination
- share a `staticVarWaiting` dependency map between parent and child query contexts
- detect circular static-variable dependencies across:
  - same or nested query contexts
  - sibling contexts waiting on variables evaluated by each other
- make static variable values `volatile` so cached results can be read before entering synchronization
- add `QueryException.Data` and `InputInfo.copy()` so errors from failed static-variable evaluation can be reconstructed for waiting threads without sharing one mutable `QueryException` instance

Regression tests are added for #2678, including:

- concurrent access to main-module and imported-module static variables
- lazy static variables evaluated only once
- side effects from lazy evaluation fired only once
- circular dependencies across sibling fork-join branches
- nested fork-join self-dependency
- copied error reporting for concurrent circular failures


basex-core/src/main/java/org/basex/util/InputInfo.java
```diff
@@ -129,6 +129,19 @@ public void internal(final boolean value) {
     internal = value;
   }
 
+  /**
+   * Creates a copy of this instance. All state is preserved, including lazy-initialization data,
```
ChristianGruen: @GuntherRademacher A »lazy« question… What would happen if we didn’t copy the InputInfo instances?

ChristianGruen:
If the evaluation of a static variable yields an error, it cannot be caught:

```xquery
(: raises FOER0000 :)
declare variable $x := error();
try { $x } catch * {}
```

So I assume that all the new concurrent states are mostly required to handle the circular dependencies. Which feels a bit over the top for a rare use case.

Maybe we find a simpler solution, possibly by allowing only one evaluation of static variables at the same time, or by replacing an invalid result by an `fn:error` instance (something we do, for example, for `if(<a/>/a) { 1 + '' }` at compile time).

basex-core/src/main/java/org/basex/util/InputInfo.java
GuntherRademacher: Reusing the same instance in multiple threads would create a race condition when `line()` or `column()` gets called, and it might result in incorrect line/column values shown in error messages.

GuntherRademacher:
Agreed - an error resulting from the RHS expression of a static variable cannot be caught (per [XQuery 4.0, §4.20](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-try-catch)). Storing the error data is still necessary to guarantee consistent behavior across multiple concurrent threads attempting to evaluate the same variable. Re-evaluating in case of error would be insufficient, if the variable's expression is nondeterministic. An example would be:
```xquery
  declare %basex:lazy variable $x := if(random:double() < 0.5) then error() else 42;
  xquery:fork-join((1 to 2) ! fn() { $x })
```
Without storing the error, if the first thread evaluates and errors, the second thread would re-evaluate with a fresh random value, potentially succeeding and caching 42. One thread saw an error, the other sees 42: two different outcomes for the same variable, which must be prevented.

The remaining code is for serializing concurrent evaluations and detecting circular dependencies before they deadlock, which may happen when fork-join branches race to evaluate the same or mutually dependent variables.

GuntherRademacher:
Concerning catchability: `Try` had a pre-existing bug: the `isCatchable` check existed in `optimize` but was absent in `value`, so non-catchable errors passed through `matches` and were caught at runtime. The fix (3f0941b) moves the check into `matches`, making it the single point to check for both callers. Tests added to `XQuery4Test.tryy`.

GuntherRademacher:
In the first attempt, exceptions were completely reconstructed, including the XQuery execution stack. The latter is however dispensable for those reproductions. The latest commit (9de3442) simplifies reproduction, eliminating `QueryException.Data` and `InputInfo.copy`. This significantly reduces the size of the changes in this PR.

ChristianGruen:
Merged (29b41c7b17cf38def548595c82a1af31c31e30e2).

Issue #2673, opened 2026-05-25
Non-determinism: counts


ChristianGruen:
Non-deterministic code should not be optimized  away. Code like…

```xquery
count($ten-strings ! (file:delete(.), .))
```

…is rewritten to `count(replicate("", 10))`.

A single `count(file:delete('x'))` will delete the file.

ChristianGruen:
I removed the bug label. Before implementing a solution, we should address two questions:

1. Consistency: how many other cases are there in which non-deterministic execution is skipped due to empty/exists/count/… calls?
2. Performance/Usability: Could enforcing non-deterministic execution negatively affect quick estimations of the expected result size?

GuntherRademacher:
> ... we should address two questions:

1. Consistency: a search for possible loss of execution of nondeterministic code execution brought up a fair amount of optimization that need to be enhanced with NDT checks. The fixes are now covered in #2695.
2. Performance: for expressions flagged as non-deterministic, performance may indeed be inferior, because some quick size shortcuts and early termination must be given up. Otherwise `count`, `empty`, or `util:count-within` may return the right boolean/count while silently skipping side effects.

ChristianGruen:
You convinced me soon that we need a more consistent solution. We already have many simple cases in which counts are performed if the code to be evaluated is nondeterministic (for example, `exists(trace(1))`), so we should certainly not deviate from this pattern in arbitrary indidivual cases.

Thanks for the PR; I’ll spend some more time on it soon.

PR #2695, opened 2026-06-17
Add missing NDT checks

GuntherRademacher:
Following up on #2673, this PR investigates optimizer rewrites that can skip execution of nondeterministic code and thereby lose its side effects.

The review found roughly 20 distinct rewrites with this problem, covered by about 30 tests. Most were fixed by adding the missing `Flag.NDT` checks. A few needed more than that, because nondeterminism is not always visible as a flag on the operand:

- a pre-evaluated function item or closure is a value and does not carry `Flag.NDT`, so some rewrites now consult the function's own nondeterminism (e.g. in `DynFuncCall` and, for the `for`-to-`replicate` rewrite, via a deferral in `GFLWOR`),
- a function invoked through a not-yet-resolved variable (`$f()`) carries no flags of its own. A new `Expr.mayInvokeVariable()` check lets the affected rewrites treat such a call conservatively until the variable is replaced by the actual function.

All affected rewrites carry comments explaining the check. Tests have also been added, each test covers one or more fixes and documents the relationship in its Javadoc. Note that these relationships can vary depending on whether the patch is applied to a fully fixed source tree or to an otherwise unfixed one.

Care was taken not to block optimizations of deterministic code. The size and count rewrites still apply to deterministic input, and where a guard had to be conservative (a function reached through a not-yet-resolved variable), the restriction is lifted again as soon as the variable is replaced by a deterministic function - so deterministic code is optimized as before. The changes are otherwise scoped to expressions marked with `Flag.NDT`.

The tests are collected in the new `NonDeterministicTest` class. The previous `RewritingsTest.nonDeterministic` test has been moved there as `NonDeterministicTest.pragma`.

See the test Javadocs and fix comments for more details.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/CompileContext.java
+++ b/basex-core/src/main/java/org/basex/query/CompileContext.java
@@ -376,7 +376,10 @@ public final class CompileContext {
   public Expr replicate(final Expr expr, final Expr count, final InputInfo info)
       throws QueryException {
     final ExprList args = new ExprList().add(expr).add(count);
-    if(expr.has(Flag.NDT, Flag.CNS)) args.add(Bln.TRUE);
+    // force repeated evaluation for side-effecting input and for calls through an unresolved
+    // variable that may be bound to a nondeterministic function. If the variable is later replaced
+    // with a deterministic function, FnReplicate drops the flag again, so nothing is pessimized.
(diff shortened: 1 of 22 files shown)


ChristianGruen:
Merged (3ccb477ada1b15835d3c42ac6439c005017dbb57), some checks added (504f7a3cd84aad163adaf13cdf97f65ec0204800).

PR #2698, opened 2026-06-20
Fix `IndexOutOfBoundsException` for non-existing groups in `fn:replace`

GuntherRademacher:
As reported by [Amanda Galtman on basex-talk](https://mailman.uni-konstanz.de/mailman3/hyperkitty/list/basex-talk@mailman.uni-konstanz.de/thread/VGVQSVY7RET2VM5VAX5SKIJZDBTY3DDL/), an `IndexOutOfBoundsException` can occur when the replacement string contains a reference to a non-existing capture group that is preceded by a backslash, e.g.

```xquery
replace('full stop.','\.','\\$1')
```

The issue is caused by `FnReplace.item`, which only examines the single character preceding a dollar sign. If that character is a backslash, the dollar sign is assumed to be escaped, and the replacement string is subsequently passed to `Matcher.replaceAll`.

The fix is to count the number of consecutive preceding backslashes. The dollar sign is escaped only if that count is odd.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/func/fn/FnReplace.java
+++ b/basex-core/src/main/java/org/basex/query/func/fn/FnReplace.java
@@ -88,7 +88,7 @@ public final class FnReplace extends RegExFn {
           if(i == -1) {
             sb.append(string, s, sl);
             s = sl;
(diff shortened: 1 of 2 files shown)


ChristianGruen:
Thank you

PR #2699, opened 2026-06-24
More NDT checks

GuntherRademacher:
Some work on a different topic brought up a few more places where nondeterministic input could be dropped, which are fixed here, in
- `fn:exactly-one`
- `fn:one-or-more`
- `fn:zero-or-one`
- `fn:tail`
- `fn:trunk`


ChristianGruen:
A quick NDT analysis resulted in more than 50 other cases that we could look at.

Before proceeding further, we may need to define which cases we want to catch or ignore. The fact that we don’t adhere to consistent rules when compiling and evaluating code complicates matters further. It may happen that a function argument is evaluated at runtime even though it will not be required for computing the result, whereas we drop it at compile time (`string-join(<a/>[. = 'x'], trace('X'))` vs. `string-join(<a/>, trace('X'))`) …and vice versa.

PR #2701, opened 2026-06-27
Fix syntax error for `URIQualifiedName` followed by a colon

GuntherRademacher:
As reported by [Andrew Sales on basex-talk](https://mailman.uni-konstanz.de/mailman3/hyperkitty/list/basex-talk@mailman.uni-konstanz.de/thread/CCRX3IDP7HJAJGLMTQ4QYPL7BWNQV2MY/), a `URIQualifiedName` without the optional prefix raised a syntax error when immediately followed by a colon (e.g. the `:=` operator) with no intervening whitespace, as in:
```xquery
let $Q{}foo:=1 return $Q{}foo
```
When parsing an `EQName`, `QueryParser.eQName` speculatively consumes a `:` to check for the optional-prefix form `Q{uri}prefix:local` (qtspecs#2227). When no valid local name follows, the colon was not given back, so the surrounding `:=` could no longer be recognized. The parser now restores its position in that case.

### DIFF
--- a/basex-core/src/main/java/org/basex/query/QueryParser.java
+++ b/basex-core/src/main/java/org/basex/query/QueryParser.java
@@ -4395,16 +4395,20 @@ public class QueryParser extends InputParser {
    */
   private QNm eQName(final byte[] ns, final QueryError error) throws QueryException {
     // parse URIQualifiedName
-    final int p = pos;
+    int p = pos;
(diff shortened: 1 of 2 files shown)


ChristianGruen:
A literal hotfix… and a new [snapshot](https://files.basex.org/releases/latest/).

Issue #2141, opened 2022-09-26
XQuery Full-Text: window expression, multiple hits, stemming


ChristianGruen:
```xquery
'a b b' contains text 'a b' all words window 2 words
```

Expected: `true`, returned: `false`.

ChristianGruen:
Related: `'a b c b' contains text 'a b c' all words ordered`

ChristianGruen:
**Expected:**

```xquery
'adverse' contains text 'adverse' using stemming,
'fox' contains text 'fox' any word using stemming
```

**Unexpected:**

```xquery
'adverse' contains text 'adverse' all words using stemming,
'adverse' contains text 'adverse' any word using stemming
```

ChristianGruen:
Discarded; too specific.

PR #2702, opened 2026-06-27
2141

dominikk:
[FIX] full-text: all words/any word window and ordered now handle rep…eated query words [#2141]

[FIX] full-text: avoid double-stemming of query terms in all words/any word [#2141]


ChristianGruen:
Considered in 8378bf57537173509d8f56da0581bc674e56013c and 6bc2361acaa1e8cbbc2c667417c48ee1b129a871; thanks.

PR #2708, opened 2026-06-30
JSON parsing: add streaming for DIRECT, W3_XML, ATTRIBUTES formats

### LINKED ISSUE #2655: JSON Parsing: Large files
This one is certainly a bit more complex as we have multiple formats for parsing JSON, all of which would probably need to be supported (or at least checked):

---------

The following command script raises an array size error:

```xml
SET PARSER json
SET CREATEFILTER *
CREATE DB test https://data.dnb.de/opendata/authorities-gnd-kongress_lds.jsonld.gz
```

Large JSON input should be streamed when creating new database instances.

GuntherRademacher:
This adds a streaming JSON parser path for database builds, avoiding the intermediate in-memory XML tree for JSON-to-XML conversion.

Streaming is now used for the `direct`, `attributes`, `w3-xml`, and deprecated `basic` formats. The existing conversion path is retained for `jsonml` and for `merge=true` direct/attributes conversions, where the output depends on whole-document type aggregation.

The change adds builder-backed JSON converters and regression coverage comparing streaming output with the existing non-streaming converters, including duplicate handling, W3 XML namespace/name compatibility, parse errors, and the `merge=true` fallback to the non-streaming path.

With these changes, the `CREATE DB` mentioned in #2655 can be executed successfully. Please close #2655 if you consider these changes sufficient, otherwise please add any requirements that should be met in addition.

### DIFF
--- a/basex-core/src/main/java/org/basex/build/Parser.java
+++ b/basex-core/src/main/java/org/basex/build/Parser.java
@@ -144,7 +144,7 @@ public class Parser extends Job {
     final MainParser mp = options.get(MainOptions.PARSER);
     final SingleParser p = switch(mp) {
       case HTML -> new HtmlParser(source, options, options.get(MainOptions.HTMLPARSER));
-      case JSON -> new JsonParser(source, options, options.get(MainOptions.JSONPARSER));
+      case JSON -> JsonStreamingParser.get(source, options);
       case CSV -> new CsvParser(source, options, options.get(MainOptions.CSVPARSER));
       default -> options.get(MainOptions.INTPARSE) ? new XMLParser(source, options) :
         new SAXWrapper(source, options);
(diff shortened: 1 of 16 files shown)


GuntherRademacher:
The second commit in this PR (be4ad85) is a refactoring that simplifies the skipping logic for duplicates under `'duplicates': 'use-first'` by moving it from the converters to the parser. Previously converters were called with `openPair(key, false)` for duplicates, and each converter individually took care of skipping. Now the skipping already happens in the parser, omitting the `add` parameters of `openPair` and `closePair`. The only converter that did not skip upon `add == false` was `JsonMLConverter` - it ran its body unconditionally, relying on `atts.add()` to detect and reject duplicates. To preserve that behaviour under the new scheme, where the parser calls `skipValue()` before ever reaching the converter, `duplicates` is now forced to `reject` for the `JSONML` format. Finally, some common methods have been moved to superclass `JsonBuilderConverter`.

ChristianGruen:
Perfect. I have adopted the code unchanged and I have separated streaming and construction in a subsequent step (83ac2f9e2406614e2c57276294625fab550f64da).

PR #2726, opened 2026-07-22
Inspect base items of large sequences when checking nondeterminism

GuntherRademacher:
This is a follow-up to the recent atomic-type check in `DynFuncCall.containsNdtFunction`, which still walked every item of large non-atomic sequences and arrays, e.g.
```
array { (1 to 100_000_000_000) ! identity#1 }(1)
```

The check now inspects the compact base items instead: a new `Value.baseItems()` (overridden in `SingletonSeq` and `SubSeq`) collapses `fn:replicate` sequences, and arrays are inspected via an `ItemArray`'s backing or a multi-member array's members (never flattened).

Regression tests have been added to `NonDeterministicTest`.


ChristianGruen:
I have chosen a slightly different fix.

PR #2741, opened 2026-08-11
DBA: language selection and result-view improvements

GuntherRademacher:
This is a follow-up to the CodeMirror 6 editor in the DBA.

Editor
- language follows the file extension: XQuery (default and .xq*), XML, JSON, otherwise plain text
- no syntax-error marker on an empty editor

Result view (read-only)
- highlighting follows the result content (XML / JSON / text)
- code folding

Bundle
- rebuilt `static/cm6.js`: adds JSON and a runtime language switch
- refreshed XQuery function signatures and documentation links
- backward-compatible

The read-only Jobs result viewer gains the same folding, other pages are unchanged.


ChristianGruen:
I’m afraid this one will need to be rebased, if possible at all: `js.js` has been split into many single files, and there is one central place now where xquery suffixes are checked.

GuntherRademacher:
Merged branch 'main' into 'cm6', adapting editor/result language selection to the restructured DBA ('files.js').

PR #2743, opened 2026-08-11
`Options`: cache option definitions per class

GuntherRademacher:
Option definitions are declared as static fields, so they are identical for every instance of an `Options` subclass. 

With the changes in this PR, they are now reflected and assembled once per class, in a `ClassValue` cache, instead of on every instantiation. Only the per-instance value map is cloned.

This speeds up construction of all `Options` subclasses. E.g. `MainOptions` instantiation is now 11 times faster, with 5 times less allocation - this was a hot spot on parse-heavy workloads.

### DIFF
--- a/basex-core/src/main/java/org/basex/util/options/Options.java
+++ b/basex-core/src/main/java/org/basex/util/options/Options.java
@@ -60,8 +60,44 @@ public class Options implements Iterable<Option<?>> {
   /** Comment in configuration file. */
   private static final String PROPUSER = "# Local Options";
 
-  /** Map with option names and definitions. */
-  private final TreeMap<String, Option<?>> definitions;
+  /**
+   * Cached option metadata (identical for all instances of a subclass).
+   * @param all all options in declaration order (with comments)
+   * @param definitions option names to definitions
+   * @param values default values (cloned per instance)
+   */
+  private record Meta(Option<?>[] all, SortedMap<String, Option<?>> definitions,
+      TreeMap<String, Object> values) { }
+
+  /** Metadata cache, computed once per subclass. */
+  private static final ClassValue<Meta> META = new ClassValue<>() {
+    @Override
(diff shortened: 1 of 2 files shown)


ChristianGruen:
Before optimizing this, it might be interesting to find why the MainOptions was particularly slow, and why so many instances needed to be created.

GuntherRademacher:
The time is spent by building definitions and values based on reflection for every instance. The per-instance cost is proportional to the number of options, `MainOptions` has 65.

The changes now do the same preparation work once per option class, share the immutable definitions, and clone the map of default values at instantiation time.

Instantiating `MainOptions` took 7.5 µs, and it goes down to 0.8 in the changed version, measured like this:

```java
  public static void main(final String[] args) {
    final int n = args.length > 0 ? Integer.parseInt(args[0]) : 1000000;
    final long t0 = System.nanoTime();
    for(int i = 0; i < n; ++i)  new MainOptions(false);
    final long t1 = System.nanoTime();
    System.out.println((t1 - t0) / n / 1000.0 + " µs per MainOptions instance");
  }
```

Each invocation of `fn:parse-xml` creates two instances of `Options`: 
- `ParseXmlOptions` (10 options)
- `MainOptions` (65 options)

Instantiating them took a considerable amount of time when parsing small documents: the following query took 17.9 µs per parse, and it goes down to 9.7 µs in the changed version:

```xquery
declare variable $DOCS external := 1000000;

let $options := map { 'intparse': true() }
let $t0 := prof:current-ns()
let $checksum := sum(
  for $i in 1 to $DOCS
  let $doc := "<doc><a></a><b>" || $i || "</b></doc>"
  return count(parse-xml($doc, $options)//node())
)
let $t1 := if($checksum ge 0) then prof:current-ns() else 0
return ($t1 - $t0) div $DOCS div 1000 || " µs per parse"
```

PR #2747, opened 2026-08-28
Fix positional comparisons with empty bounds

GuntherRademacher:
Some XQuery code incorrectly resulted in 
```txt
[XPTY0004] Number expected, item() found: ().
```

for a general comparison with an empty operand. This can be reproduced by
```xquery
let $x := 0
let $first-nonzero := trace(head(for $c at $i in $x where $c!=0 return $i), "first-nonzero")
return $x[position() >= $first-nonzero]
```

It is caused by positional comparisons being rewritten to `util:range` regardless of the empty argument.

### DIFF
--- a/basex-core/src/test/java/org/basex/query/expr/ExprTest.java
+++ b/basex-core/src/test/java/org/basex/query/expr/ExprTest.java
@@ -651,6 +651,15 @@ public final class ExprTest extends SandboxTest {
     // GH-2224: Unexpected exception, arithmetic operations with positional expression
     check("document { <X/> }//X[not(position() * 2 = last())]", "<X/>");
     check("document { <X/> }//X[not(position() + position() = last())]", "<X/>");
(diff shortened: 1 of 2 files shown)


basex-core/src/test/java/org/basex/query/expr/ExprTest.java
```diff
@@ -651,6 +651,15 @@ public final class ExprTest extends SandboxTest {
     // GH-2224: Unexpected exception, arithmetic operations with positional expression
     check("document { <X/> }//X[not(position() * 2 = last())]", "<X/>");
     check("document { <X/> }//X[not(position() + position() = last())]", "<X/>");
+
+    // GH-2747: Avoid rewriting positional range predicates to util:range if bounds may be empty
```
ChristianGruen, line 655: …correct ;)

Issue #2763, opened 2026-09-16
`fn:invisible-xml`: make it interruptible


ChristianGruen:
We should make `fn:invisible-xml` interruptible. With the grammar `s: s, s | "a".` and just 20 characters of input `aaaaaaaaaaaaaaaaaaaa`, the JVM runs out of heap. Limiting the input size doesn't help, so an ambiguous grammar can crash the whole server.

An example:

```xquery
xquery:eval(
  'declare variable $grammar external;
   declare variable $input external;
   invisible-xml($grammar)($input)',
  { 'grammar': 's: s, s | "a".', 'input': 'aaaaaaaaaaaaaaaaaaaa' },
  { 'timeout': 2, 'memory': 512 }
)
```

PR #2767, opened 2026-09-22
Make `fn:invisible-xml` parser interruptible

GuntherRademacher:
As reported in #2763, an ambiguous ixml grammar can grow the parse forest until OOM without being stoppable: `xquery:eval`'s `timeout` and `memory` limits never fire, because the parse loop does not take notice of the `stopped` condition.

Markup Blitz 1.13 now has a checkpoint-aware `parse`. Passing `qc::checkStop` to it makes these limits work.


ChristianGruen:
Great

